PR #10, opened 2018-04-05
Ignore _book directory

zimeon:
Add `_book` to `.gitignore`


zimeon:
I added an ignore for `/book.pdf` which is what you get if building a one-file copy of the spec with `gitbook pdf`

Issue #18, opened 2018-05-25
Move gist contents to the spec for commenting, etc.


ahankinson:
n/t

awoods:
In the rendition of the gist that will appear in the specification, please annotate which elements are required vs. the ones that are included as example content.

awoods:
Resolved with: https://github.com/OCFL/spec/pull/20

PR #20, opened 2018-05-30
Added an example minimal structure

ahankinson:
This PR adds a minimal OCFL object structure to the specification.

Refs #18


julianmorley:
+1

Issue #9, opened 2018-04-04
What is an OCFL digital object?


zimeon:
We need a good definition for the [glossary](https://ocfl.github.io/spec/GLOSSARY.html#digital-object) (https://github.com/OCFL/spec/blame/master/GLOSSARY.md#L2 ).

zimeon:
I think we should not try to describe an object in the abstract but instead do it in terms of the OCFL properties. For a first stab, could be something along the lines of:

> An OCFL digital object is a collection of zero or more datastreams and administrative information that together have a URI identifier. An object may contain a sequence of versions of the datastreams.

ahankinson:
Do you think the definition in the discussion paper provides a good starting point?

zimeon:
My comment was motivated by thinking the definition in the paper is not adequate:

> An OCFL digital object is a collection of files and metadata that can have a notional but largely implicitly-understood boundary: “This is the thing, and this is not the thing.” ... The object should also contain a record of the metadata that describes the origin, character, and purpose of the collection of files. ...

because: 

  * I think an OCFL digital object has a very well defined boundary and, like BagIt, a notion of testable completeness  (how that maps to boundaries in other models _is_ arbitrary, but out of scope)
  * It omits the idea of having an identifier
  * I think the description of what the metadata describes is an overreach (it should enable all of those things to be described, but they are note required.

Looking at it now, I think my https://github.com/OCFL/spec/issues/9#issuecomment-378752652 was excessively reductionist, and omits both the metadata/data distinction and log files

julianmorley:
Is OCFL actually an AIP? Or is it a set of standards that other AIP formats can conform to, to gain compatibility and extended features? For example, what does an OCFL-compliant Moab object look like? What does an OCFL-compliant Bagit object look like? What benefits (aside from cross-compatibility) do they gain from compliance?

I know that I am putting myself on the hook to answer at least some of these questions.

ahankinson:
I believe yes, it is a standardised form of an AIP.

rosy1280:
@julianmorley its a standard form of an AIP

from @rotated8 its a grouping of files that represent an intellectual endeavor

ahankinson:
> An OCFL Object is a collection of one or more bitstreams and administrative information that together have a URI identifier. An object may contain a sequence of versions of the datastreams.

ahankinson:
> OCFL is a specification that describes how OCFL Objects behave in a structure, transparent, and predictable manner.

[...]

PR #25, opened 2018-06-20
A stab at definitions for effort and object 

zimeon:
Based on discussion in #9 with some opinion thrown in ;-)


ahankinson:
Pinging @neilsjefferies

Issue #29, opened 2018-07-11
Every inventory file MUST have an accompanying digest file


zimeon:
In discussion https://github.com/OCFL/spec/wiki/2018.07.11-Editors-Meeting we have agreed that each `inventory.jsonld` file MUST have an accompanying digest file `inventory.jsonld.DIGEST` where `DIGEST` may be `sha512` or such, whether this be at the top level or inside a version.

Issue #21, opened 2018-05-31
Allow for user-determined checksum algorithms 


SibylSchaefer:
Currently, SHA-512 is used only in a minority of preservation settings as the fixity checking algorithm (see the 2017 [NDSA Fixty Survey Report](https://ndsa.org/documents/Report_2017NDSAFixitySurvery.pdf), p22 -23). Requiring SHA-512 may be a cumbersome barrier for those who want to adopt OCFL.

ahankinson:
Thanks, Sibyl! And especially for the link to the survey report -- very interesting!

A brief summary of the discussions on this issue to date, since they don't seem to be captured anywhere:

The choice of SHA-512 should be viewed as an alternate to SHA-256, which we discussed in issue #8 .  It turns out that SHA-512 is faster to compute, so would be less of a burden.

In the first community call we discussed the choice of Moab to support both MD5 and SHA1. A comment was made by someone on that call that they wished it just supported a single hashing algorithm, for simplicity of implementation and checking.

One of the principle functions of the hashing algorithm is to provide content-addressable storage and de-duplication. It may also be used as a fixity check, but the choice of hashing algorithm should be understand in light of providing both functions.

MD5 and SHA1 are widely used, but are not necessarily future-proof as they are both now cryptographically [insecure](https://shattered.io).

Someone suggested to me that we adopt Multihash values: https://github.com/multiformats/multihash. I'm a bit wary of this as it is an external dependency to our work. 

We could allow specification of the hash algorithm on the object level, but we won't be able to mix hash types; you can't have half your hashes as MD5, and the other half as SHA1.

ahankinson:
(My personal view is that we should use OCFL to drive best-practice adoption, which would suggest the use of SHA-512 by default, but we may be able to support SHA1 and MD5 for 'legacy' content.)

julianmorley:
I concur with @ahankinson - sha512 by default, with an option to 'downgrade' on a per-object basis to another algo (or upgrade, as the spec evolves). 

And I think it was me who was lamenting about the many checksums in Moab (and if it wasn't, I'd add my lament anyway). The reasoning for Moab was that it provides an additional defense against checksum collisions and potential malicious manipulation of files that, after manipulation, produce the same checksum but are clearly no longer the same content. It also provides a more seamless migration of existing content to new algorithms - just drop the oldest and add the new one, whilst maintaining the 3rd algorithm as a common bridge. The viability and necessity of that approach is definitely open for discussion.

[...]

PR #34, opened 2018-07-18
Fixed: Add section on inventory sidecar files

ahankinson:
This PR adds a section on hash sidecar files.

Refs #29

### DIFF
--- a/index.html
+++ b/index.html
@@ -209,6 +209,21 @@
     <h2>Digest Algorithm Choice</h2>
   </section>
 
+  <section id='digest-sidecar-file'>
+    <h2>Digest Sidecar File</h2>
+    <p>
+        Every occurrence of an inventory file MUST have an accompanying sidecar file stating its checksum. This sidecar file must be of the form <pre>inventory.jsonld.HASHALG</pre>, where <pre>HASHALG</pre> follows the chosen hashing
+        algorithm for the object. An example might be <pre>inventory.jsonld.sha512</pre>.
+    </p>
+    <p>
+        The digest sidecar file MUST contain the hash of the inventory file. This MUST follow the format <pre>CHECKSUM inventory.jsonld</pre>; that is, the hash of the inventory file, a single space, and the name of the file being checksummed.
+    </p>
+    <p>
+        The hash of the inventory MUST be computed only after all changes to the inventory 
+        have been made. As such, it is RECOMMENDED that this sidecar file be created as the last step in creating a new version.
+    </p>
+  </section>
+
   <section id='object-manifest'>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -209,6 +209,21 @@ <h2>Basic Structure</h2>
     <h2>Digest Algorithm Choice</h2>
   </section>
 
+  <section id='digest-sidecar-file'>
+    <h2>Digest Sidecar File</h2>
+    <p>
+        Every occurrence of an inventory file MUST have an accompanying sidecar file stating its checksum. This sidecar file must be of the form <pre>inventory.jsonld.HASHALG</pre>, where <pre>HASHALG</pre> follows the chosen hashing
```
awoods: Is there a canonical list of hash algorithms we can reference?

zimeon: See comments on #21 -- I think the answer is a bit tricky but we will need to define the list, hopefully by reference to one or more other specs. I would suggest we merge this and solve that as a separate problem

zimeon:
Is the addition of the filename following parallel with the bagit approach? I don't object to that but it seems unnecessary

awoods:
@zimeon : https://tools.ietf.org/html/draft-kunze-bagit-08#section-2.1.3

zimeon:
Yes, I see it there and there it is necessary because the `manifest-XYZ.txt` file contains checksums for multiple files. However, we have a different naming convention and the checksum for just one file... so it isn't really necessary

ahankinson:
It is also used in the output of `shasum`:

```
$>shasum -a 512 CNAME
dba3b3255390e9bda6fcc1f3cc8534a4c9fec4f90478c8e62bfaae73fab5700d524a8fb09d692c41cdfd83ec978a282082b586392325124c7cd5c21cca115599  CNAME
```

and is also used in software distributions as sidecars, e.g.,

http://www-eu.apache.org/dist/lucene/solr/7.4.0/

(although I'm not sure what the * in the filename means).

zimeon:
Thanks @ahankinson - I agree the * is odd but good to see similar use in other places

ahankinson:
I updated per #35 so re-review and approval is necessary.

ahankinson:
@zimeon This section in the output of the `shasum` utility explains what the * is for:

> The sums are computed as described in FIPS-180-4.  When checking, the
input should be a former output of this program.  The default mode is to
print a line with checksum, a character indicating type (`*' for binary,
` ' for text, `?' for portable, `^' for BITS), and name for each FILE.

So maybe we also need to decide the mode?

zimeon:
Will port your comment above onto #39

PR #42, opened 2018-07-19
Fixed: Formatting, Language

ahankinson:
This PR adds two non-substantive changes:

1. Normalizes the file to a consistent use of spaces (4 per indent) and removes tabs. Also converted to a consistent use of quotation marks.
1. Updates the "Versions Directories" section to use RFC 2119 "SHOULD" instead of "RECOMMENDED."

### DIFF
--- a/index.html
+++ b/index.html
@@ -198,7 +198,7 @@
             integers: 1, 2, 3, etc., and the version directory name is constructed by adding the prefix <code>v</code>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


awoods:
@ahankinson : Looks like this PR has merge conflicts.

ahankinson:
Since a) it's non-substantive, and b) it looks like we have quorum on approval, I will re-do the changes and push directly to master, which is easier than trying to merge it.

Issue #30, opened 2018-07-11
Should OCFL be defined in terms of files or bitstreams?


zimeon:
The current definition of an [OCFL object](https://ocfl.io/#object-spec) says:

> An OCFL Object is a group of one or more content bitstreams (data and metadata), and their administrative information that are together identified by a URI. The object may contain a sequence of versions of the bitstreams that represent the evolution of the object's contents.

If the name of the initiative/spec includes file in the title, should we include file in the definition of an OCFL object? If so, how do we indicate that this work can be generalized to work over object stores too?

awoods:
I can see using "file" in the above definition, then having a footnote (or subsequent definition) stating that a "file" can include any persisted bitstream.

zimeon:
:+1: - I think that would be a reasonable resolution of the issue

zimeon:
I note that this while talking about file<->bitstream we might also mention directory<->path-segment

zimeon:
Addressed by #44 and #97, also related to #28

Issue #28, opened 2018-06-22
Object specification definition - revision suggestion


peterVG:
We discussed the definition for Objects in today's community call. It was suggested to move this to GH as a next step. I'm not able to branch and pr on the repo atm so I will post this as an issue here instead. The following captures the changes I had suggested to put the emphasis back on the concept of "files" in the primary specification definition. I'll leave it to the spec editors to do with as you please:

3. Object Specification
"An OCFL Object is a group of one or more content and metadata files (e.g. file1.jpg, file2.txt, file3.ead.xml, file4.dc.json). This object is identified by an URI and includes OCFL administrative information. This administrative information may contain a sequence of versions of the files that document the evolution of the object's contents. The files in the object follow a layout as prescribed by the OCFL specification."

SibylSchaefer:
Regarding the last sentence: is it the files in the object that follow the OCFL layout, or the versions of the object which do? My understanding was the latter.

peterVG:
My understanding from reading the OCFL background paper is that it is both. E.g. "Object versioning is implemented using a folder hierarchy within the digital object’s root directory"
https://docs.google.com/document/d/13gFfSu1fePKx0eQYk458zE6OZwCW5JH_hTIG58f3S8c/edit) The way I read it, OCFL specifies how to include versions of an object as well as all other content, metadata, and administrative information (eg. .logs, versions.jsonld) about that object. So maybe the  last sentence could be changed to "The files in the object, along with information about any of its versions, follow a layout prescribed by the OCFL."
I am not authenticated to make PRs in this repo so I'll leave it to someone else to put the request in formally if it is deemed worthwhile. I am kind of parachuting into this existing territory so I certainly don't want to make any unsubstantiated comments that unsettles prior work. I was just following up on the request to document the discussion as an interested third-party and the comment on the role of "files" in the definition.

zimeon:
My understanding of the reason to talk about bitstreams rather than files was to make it clearer that we cover things in object stores (such as S3) rather than just filesystems. Having said that, one can either write a general definition and then later talk about implementation on a filesystem or object store, or could write a slightly over-specific definition and then generalize later by says it could also work on object stores.

Either way, I think it would be a mistake to put example file/object names, especially ones with some suggested meanings based on implicit extension typing, in the definition.

peterVG:
[...]

PR #44, opened 2018-07-19
WIP: Object definition expansion

ahankinson:
Attempts to coalesce issues #28 and #30. Makes the distinction between
the 'structure' of an OCFL object, and the 'content' it contains, as per
our discussions on the editors meeting call (11.07.2018).

### DIFF
--- a/index.html
+++ b/index.html
@@ -144,10 +144,29 @@
 <section id="object-spec">
     <h2>Object Specification</h2>
     <p>
-        An OCFL Object is a group of one or more content bitstreams (data and metadata), and their administrative
-        information that are together identified by a URI. The object may contain a sequence of versions of the
-        bitstreams that represent the evolution of the object's contents.
+        An OCFL Object is a group of one or more files, directories, and administrative
+        information, that are together identified by a URI. The object may contain a sequence of versions of the
+        files that represent the evolution of the object's contents.
     </p>
+    <p>
+        A file is defined as a content bitstream that can be stored and transmitted. Directories (also called "folders")
+        allow for the organization of files into tree-like hierarchies. The content of an OCFL Object is the files and
+        directories that are stored <i>within</i> the hierarchy layout provided as part of the spec.
+    </p>
+    <p>
+        The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to
+        store their own metadata serializations as files in an OCFL Object, but, in the context of the OCFL Specification
+        these are considered to be part of the contents of an OCFL Object.
+    </p>
+    <p>
+        An OCFL Object is therefore:
+    </p>
+    <ol>
+        <li>A conceptual gathering of all files (data and metadata), folders, and their changes over time which together
+            form the digital representation of a complete intellectual endeavour (i.e., content); and</li>
+        <li>A file and folder layout and administrative information on a storage medium that provides a defined structure
+            for the storage of this content, and through which these files and their changes may be understood (i.e., structure).</li>
+    </ol>
 
     <section id="basic-structure">
         <h2>Basic Structure</h2>


index.html
```diff
@@ -144,10 +144,29 @@ <h2>Terminology</h2>
 <section id="object-spec">
     <h2>Object Specification</h2>
     <p>
-        An OCFL Object is a group of one or more content bitstreams (data and metadata), and their administrative
-        information that are together identified by a URI. The object may contain a sequence of versions of the
-        bitstreams that represent the evolution of the object's contents.
+        An OCFL Object is a group of one or more files, directories, and administrative
```
zimeon: I think there is a problem including "directories" as a fundamental component of an OCFL object because OCFL doesn't actually store directories (e.g. empty dir will be ignored). It does support "files organized in directories" so maybe that could be used as a replacement?

index.html
```diff
@@ -144,10 +144,29 @@ <h2>Terminology</h2>
...
+        information, that are together identified by a URI. The object may contain a sequence of versions of the
+        files that represent the evolution of the object's contents.
     </p>
+    <p>
+        A file is defined as a content bitstream that can be stored and transmitted. Directories (also called "folders")
+        allow for the organization of files into tree-like hierarchies. The content of an OCFL Object is the files and
+        directories that are stored <i>within</i> the hierarchy layout provided as part of the spec.
```
zimeon: Same issue with directories as above

index.html
```diff
@@ -144,10 +144,29 @@ <h2>Terminology</h2>
...
+    <p>
+        The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to
+        store their own metadata serializations as files in an OCFL Object, but, in the context of the OCFL Specification
+        these are considered to be part of the contents of an OCFL Object.
+    </p>
+    <p>
+        An OCFL Object is therefore:
+    </p>
+    <ol>
+        <li>A conceptual gathering of all files (data and metadata), folders, and their changes over time which together
```
zimeon: are we calling them directories or folders? It was directories above. We should be consistent

index.html
```diff
@@ -144,10 +144,29 @@ <h2>Terminology</h2>
...
+            form the digital representation of a complete intellectual endeavour (i.e., content); and</li>
```
zimeon: I still feel that introducing the notion of a "complete intellectual endeavour", whatever that might mean, is spurious. We also need to decide on US or UK spelling and use that consistently: endeavor or endeavour. I think I would scratch everything from "An OCFL object is therefore:" onward -- if it isn't clear from the first portion then we should fix that instead of offering a second try

index.html
rosy1280: How about instead:

> An OCFL Object is a group of one or more files, administrative information, and the directories they are organized in, that are together identified by a URI.

>

index.html
rosy1280: but isn't this section defining what a directory is? i guess i'm not sure how you can define a directory without using the word directory, or maybe i'm misunderstanding the comment (probably this).

index.html
rosy1280: My concern with not including something like a "intellectual endeavor" (US spelling or bust) is that folks aren't going to understand what to put in an OCFL object.  

We often had this argument in the early days of DPN.  Some folks felt that an object was an entire collection, others felt that it was a single book, while still others package up their entire repository.  

Contrast that with some of the staff here at my library who have argued that an object is a single file i.e. the PDF is an object and each one of the TIFs of the pages are also single objects.  The goal of OCFL -- in my mind at least -- is to say that an OCFL object is the book, not a collection of books nor the individual files that make up the book.

I realize this is flame war worthy, but like using US spelling, I'm willing to duke it out because I think this is an important conversation to be had.

index.html
zimeon: I think this still has the problem that it implies directories are an element of content whereas we support them only in so much as they are where the files live. I propose following @rosy1280's suggested construction above...

> The content of an OCFL Object is the files and the directory hierarchy they are organized in.

though that is pretty much redundant with the definition above. Maybe just the first two sentences of @ahankinson's proposal are enough?

index.html
neilsjefferies: I don't think we should be too proscriptive about what is in an OCFL object - as long as it has enough metadata to describe itself then I don't see why there is a problem. In particular, I want to store people, places and collections as distinct objects. What about WARC files or disk images?

zimeon: Not to get us off into the weeds further, if we were to draw a parallel with PREMIS (http://www.loc.gov/standards/premis/v3/premis-3-0-final.pdf - just because I happen to look at that tody) then I think an OCFL object would map to a PREMIS digital Representation (minus the recursion)

The point about whether you have an OCFL object for a page, a book, or a set of books is entirely up to the repository and data managers in my opinion. OCFL is completely agnostic on the issue, the object is merely a set of files organized with a directory hierarchy (or not) that are managed as one lump, with versioning

ahankinson:
Merging with caveat that this will be worked on this week.

PR #45, opened 2018-07-19
New: Begin the section on the logs directory

ahankinson:
This PR adds some text about the 'logs' directory.

### DIFF
--- a/index.html
+++ b/index.html
@@ -181,16 +181,23 @@
     <section id="logs-directory">
         <h2>Logs Directory</h2>
         <p>
-            The base directory of the object MUST contain a logs directory, which MAY be empty. Implementers SHOULD use
-            this for storing files that log any actions taken on the object. Since these logs may be subject to local
-            standards adherence needs, the format of these logs is considered out-of-scope for the OCFL Object. However,
-            clients operating on the object SHOULD log actions here that are not otherwise captured.
+            The base directory of the object MAY contain a logs directory, which MAY be empty. Implementers SHOULD use
+            this for storing files that contain a record of actions taken on the object. Since these logs may be subject
+            to local standards requirements, the format of these logs is considered out-of-scope for the OCFL Object.
+            Clients operating on the object MAY log actions here that are not otherwise captured.
         </p>
-        <p>
-            The purpose of the logs directory is to provide implementers with a location for storing local information
-            about the object that may not necessarily be part of the object itself, and to expand on the processable
-            information that can be found in the object's inventory.
-        </p>
+        <div class="note">
+            <p>
+                The purpose of the logs directory is to provide implementers with a location for storing local information
+                about actions to the object that may not necessarily be part of the object itself.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -180,6 +180,17 @@ <h2>Object Conformance Declaration</h2>
...
     <section id="logs-directory">
         <h2>Logs Directory</h2>
+        <p>
+            The base directory of the object MUST contain a logs directory, which MAY be empty. Implementers SHOULD use
+            this for storing files that log any actions taken on the object. Since these logs may be subject to local
+            standards adherence needs, the format of these logs is considered out-of-scope for the OCFL Object. However,
+            clients operating on the object SHOULD log actions here that are not otherwise captured.
+        </p>
+        <p>
+            The purpose of the logs directory is to provide implementers with a location for storing local information
```
awoods: Should this be a non-normative block?

index.html
```diff
@@ -180,6 +180,17 @@ <h2>Object Conformance Declaration</h2>
...
+            about the object that may not necessarily be part of the object itself, and to expand on the processable
+            information that can be found in the object's inventory.
```
awoods: > expand on the processable information that can be found in the object's inventory.

I am a little unclear on the meaning of the above.

neilsjefferies: Could we include an example: such as a fixity check that finds nothing wrong?

index.html
neilsjefferies: The fact that there MUST be a logs directory is normative - even if the contents are unspecified.

awoods: Agreed. The first block is normative ( https://github.com/OCFL/spec/pull/45/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR183 ), and 
the second block seems non-normative ( https://github.com/OCFL/spec/pull/45/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR190 ).

I would suggest wrapping the second block as a non-normative statement:
```
      <blockquote class="informative">
        <p>
          Non-normative note:
          The purpose of the logs directory is to provide implementers with a location for storing local information
          about the object that may not necessarily be part of the object itself, and to expand on the processable
          information that can be found in the object's inventory.
        </p>
      </blockquote>
```

index.html
rosy1280: Answer from the editors meeting: you could do it that way, which is why it can be empty.

ahankinson:
A directory named `logs` is permissible, but it's not a requirement.

ahankinson:
Changes added. Will need re-approval.

zimeon:
@ahankinson - need to fix line length to keep travis happy

ahankinson:
@awoods could you merge this please?

zimeon:
@ahankinson I think per our rules the _next_ :+1: can merge

ahankinson:
OK; It would just be nice to get this one done and merged.

awoods:
@neilsjefferies ? @rosy1280 ? @julianmorley ? are any of you available for a review/merge?

Issue #47, opened 2018-07-20
Add CI and basic validation to ReSpec


awoods:
This task is to add a .travis.yml with the following restrictions on the ReSpec document:
- [html5validator](https://pypi.org/project/html5validator/)
- line length not to exceed 120 characters

PR #48, opened 2018-07-20
Add .travis.yml

awoods:
..with the following restrictions on the ReSpec document:
- html5validator
- line length not to exceed 120 characters

Resolves: https://github.com/OCFL/spec/issues/47

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: java
+jdk:
+  - oraclejdk8
+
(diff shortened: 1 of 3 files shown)


awoods:
@zimeon : I was curious about the `java` install as well. But it appears to be needed for the underlying vnu.jar.
https://pypi.org/project/html5validator/#integration-with-travisci

PR #58, opened 2018-08-17
Fix long-line check

zimeon:
https://github.com/OCFL/spec/pull/45 wrapped the long copyright line so now we expect zero lines over 120 chars


ahankinson:
As this is a non-substantive change, I have merged it.

PR #110, opened 2018-09-07
Definition for Logical File Path

### LINKED ISSUE #85: Logical path definition
the relative file path for the version.

rosy1280:
Fixes #85


zimeon:
F2F decision: don't mention "version directory" because that is part of the existing file path and the state. Also example is confusing and best removed.

rosy1280:
F2F 2018.09.07 Also decided not to include examples in the terminology

PR #124, opened 2018-09-07
Update terms and define object root and storage root

rosy1280:


index.html
```diff
@@ -202,13 +194,9 @@ <h2>Terminology</h2>
         <dt><dfn>OCFL Object</dfn>:</dt>
         <dd></dd>
         <dt><dfn>OCFL Object Root</dfn>:<dt>
-        <dd></dd>
+        <dd>The root of an <a>OCFL Object</a>.</dd>
```
awoods: `The base directory of an OCFL Object that may exist either as a first class directory, or in the case of a zipped OCFL Object, as the top directory in a zip file`.

PR #140, opened 2018-09-07
Version definition

### LINKED ISSUE #80: Version definition
The version value MUST be the same as the directory name for the version.  Version numbers SHOULD start at 1 and MUST be positive sequential integers.

rosy1280:
Fixes #80

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -219,7 +219,7 @@
         <dt><dfn>OCFL Storage Root</dfn>:</dt>
         <dd>The base directory of an OCFL storage layout.</dd>
         <dt><dfn>Version</dfn>:<dt>
-        <dd></dd>
+        <dd>An incremental change to the preserved content of an <a>OCFL Object</a>.</dd>
     </dl>
 </section>
 
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -219,8 +219,8 @@
         <dt><dfn>OCFL Storage Root</dfn>:</dt>
         <dd>The base directory of an OCFL storage layout.</dd>
         <dt><dfn>OCFL Version</dfn>:<dt>
-        <dd>The state of an OCFL object's content which is constructed using the incremental
-          changes recorded in the sequence of corresponding and prior version directories.</a>.</dd>
+        <dd>The state of an <a>OCFL object</a>'s content which is constructed using the incremental
+          changes recorded in the sequence of corresponding and prior version directories.</dd>
     </dl>
 </section>
 
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -218,8 +218,9 @@
           directory in a zip file.</dd>
         <dt><dfn>OCFL Storage Root</dfn>:</dt>
         <dd>The base directory of an OCFL storage layout.</dd>
-        <dt><dfn>Version</dfn>:<dt>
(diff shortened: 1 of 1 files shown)


awoods:
As an adjacent point to this discussion, I believe we should make clear that "OCFL takes no position on when in the lifecycle of an OCFL Object versions should be created". https://github.com/OCFL/spec/issues/141

zimeon:
I missing something in the desire to focus on incremental change: Version 1 has no incremental change, it just is. Subsequent versions may be derived by incremental changed from v1 if that his how OCFL is used (the most likely case) but that seems an implementation detail compared with the notion of OCFL representing a sequence of versions of the whole state of the content (our fundamental abstract model)

ahankinson:
I think 'incremental' reflects the fact that we are forward-diff'ing -- 'incremental change' as opposed to 'complete change of all things in the object'

zimeon:
To me "a version" is the state and not the change --- what you might expect to get out of a system if you said "Please disseminate version 2 of OCFL Object ark:a/b/c" (e.g. [OED](https://en.oxforddictionaries.com/definition/version)). If we want to include the notion of incremental change to derive a version then perhaps:

> Version: A particular state of the OCFL object's content. A version is constructed according to the incremental changes recorded in the sequence of corresponding and prior version directories.

rosy1280:
I think the confusion may be in what we are trying to define.  There are
perhaps three things that correspond with version:

1. Version
2. OCFL Version
3. OCFL Version Directory

I was attempting to define #1, Simeon seems to be defining #2 (???) and
then of course there is #3 which I don't think anyone is trying to define
(??)

ahankinson:
Given that you reference an OCFL Object in the definition, I think it's good to assume that it's an OCFL Version (that is, essentially defining the _reason_ for the `Version` type in the `versions` array in the inventory)

rosy1280:
Then I would suggest changing the definition to:

draft/spec/index.html
```diff
@@ -466,12 +466,12 @@ <h2>Version</h2>
                             </p>
                             <p>
                                 Notably, the version state can be used to provide de-duplication of content within
-                                the OCFL Object, by mapping multiple <a>logical file path</a>s to the same content 
-                                digestin the manifest. Implementers may choose to use this functionality on a 
-                                case-by-case basis. For example, they may choose to implement renaming (that is, 
-                                changes in file <i>name</i>, but not file <i>content</i>) by pointing a new logical 
+                                the OCFL Object, by mapping multiple <a>logical file path</a>s to the same content
+                                digestin the manifest. Implementers may choose to use this functionality on a
```
awoods: typo: `digestin`

draft/spec/index.html
```diff
@@ -211,8 +211,9 @@ <h2>Terminology</h2>
           directory in a zip file.</dd>
         <dt><dfn>OCFL Storage Root</dfn>:</dt>
         <dd>The base directory of an OCFL storage layout.</dd>
-        <dt><dfn>Version</dfn>:<dt>
-        <dd></dd>
+        <dt><dfn>OCFL Version</dfn>:<dt>
+        <dd>The state of an OCFL object's content which is constructed using the incremental
+          changes recorded in the sequence of corresponding and prior version directories.</a>.</dd>
```
awoods: Travis is probably sad about the lonely `</a>`
https://travis-ci.com/OCFL/spec/builds/84563056#L466

Issue #149, opened 2018-09-11
Should we insist on lowercase hex letters in digests?


zimeon:
RFC4648 and Bagit allow either case for digests, so:

```
d41d8cd98f00b204e9800998ecf8427e
D41d8CD98F00B204E9800998ECF8427E
d41d8Cd98f00B204e9800998ECf8427e
```

are all valid as the MD5 of "".

We could follow this, we could SHOULD lowercase, or we could MUST lowercase. We should specify whichever. (Follow on from #146.)

zimeon:
I lean toward **"MUST use lowercase"**. I don't see the advantage in allowing variability and the common tools use lowercase.

awoods:
Coming from a position of not know what case ALL of the tools produce, and taking the lead from https://tools.ietf.org/html/rfc4648, I would be inclined to **"SHOULD use lowercase"**... recognizing that supporting upper or mixed case adds complexity to validation tooling.

ahankinson:
I have made a new issue for the implementation guidelines to clarify that if you are relying on string comparison for checksum identification, that you should first cast the string to one case or another. Would that be an appropriate solution?

zimeon:
So, if it is not MUST, are we OK with the following?

```
  "manifest": { "abc": [ "v1/foo.xml" ] }
  "v1" {
    "state": { "ABC": [ "v1/foo.xml" ] }
  }
```

ahankinson:
I think we have to be, as messy as that is.

awoods:
Or we could have SHOULD be lowercase and MUST be consistent across `manifest` and `state`.

ahankinson:
We could make it a warning in the validator ("Mixed case digests in inventory")

zimeon:
Regarding https://github.com/OCFL/spec/issues/149#issuecomment-420662344, there are currently no tools that produce or consume OCFL inventories so I don't think we are constrained by what existing tools do. If someone uses a digest tool/library that writes/generates hex using uppercase that could trivially be converted to lowercase within the OCFL inventory tooling

ahankinson:
I think the concern was that if the Python checksum library produced uppercase digests, and the Java one produced lowercase digests, both should be usable without needing special (normative) OCFL tooling to normalize that.

zimeon:
I don't see that as a concern, to me the tooling for dealing with OCFL inventory documents would just lowercase, e.g. Python: `digest.lower()`, Java: `digest.toLowerCase()`, etc...

[...]

PR #155, opened 2018-09-13
Added non-normative note about case comparisons

ahankinson:
My take on a solution to #149 . I think we should include a note about case insensitivity, but it should be non-normative.


zimeon:
Seems OK pending decision on #149. I would write out "lowercase" in preference to "lower-" because I think that is clearer and saving 3 characters doesn't matter much in the electronic world

PR #158, opened 2018-09-17
Definition tuning

ahankinson:
- Reformatted definitions to follow the style of the rest of the spec.
 - Reworded some definitions to make their intent clearer
 - Added missing definitions

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -204,78 +204,82 @@
     <dl>
         <dt><dfn>Existing File Path</dfn>:</dt>
         <dd>
-            The file path as it appears on disk including the version directories and the filename, and relative to
-            an <a>OCFL Object Root</a>. Existing file paths are used in the <a>Manifest</a> within an
-            <a>Inventory</a>.
+            The file path of a file on disk, relative to the <a>OCFL Object 
+            Root</a>. Existing file paths are used in the <a>Manifest</a> 
+            within an <a>Inventory</a>.
         </dd>
         <dt>
             <dfn>Digest</dfn>:
         </dt>
         <dd>
-            An algorithmic characterisation of the contents of a file conforming to a standard digest algorithm.
-            Digests are used for recording fixity and for content-addressable storage within an OCFL Object.
+            An algorithmic characterization of the contents of a file 
+            conforming to a standard digest algorithm.
         </dd>
         <dt>
             <dfn>Inventory</dfn>:
         </dt>
         <dd>
-            A file, expressed in JSON-LD, that tracks the history and current state of an OCFL Object.
+            A file, expressed in JSON-LD, that tracks the history and current 
+            state of an OCFL Object.
         </dd>
         <dt>
             <dfn>Logical File Path</dfn>:
         </dt>
         <dd>
-            A path that represents a file's location in the <a>logical state</a> of an object. Logical file paths
-            are used in conjunction with a digest to represent the file name for a given bitstream at a given version.
+            A path that represents a file's location in the 
+            <a>logical state</a> of an object. Logical file paths are used in 
+            conjunction with a digest to represent the file name for a given 
+            bitstream at a given version.
         </dd>
         <dt>
             <dfn>Logical State</dfn>:
         </dt>
         <dd>
-            A grouping of logical file paths reflecting the state of the content for a given version.
+            A grouping of logical file paths reflecting the state of the 
+            content for a given version.
         </dd>
         <dt>
             <dfn>Logs Directory</dfn>:
         </dt>
         <dd>
-            A directory for storing information about the content (e.g., actions performed) that may not
-            necessarily be part of the content itself.
+            A directory for storing information about the content (e.g., 
+            actions performed) that may not necessarily be part of the content 
+            itself.
         </dd>
         <dt>
             <dfn>Manifest</dfn>
         </dt>
         <dd>
-            A complete listing of all files, with their digests, within an OCFL Object and stored in the
-            <a>Inventory</a>.
+            A section of the <a>Inventory</a> listing all files and their digests within an OCFL Object.
         </dd>
         <dt>
             <dfn>OCFL Object</dfn>:
         </dt>
         <dd>
-            A group of one or more content files and administrative information, that are together identified by a
-            URI. The object may contain a sequence of versions of the files that represent the evolution of the
-            object's contents.
+            A group of one or more content files and administrative 
+            information, that are together identified by a URI. The object may 
+            contain a sequence of versions of the files that represent the 
+            evolution of the object's contents.
         </dd>
         <dt>
             <dfn>OCFL Object Root</dfn>:
         <dt>
         <dd>
-            The base directory of an <a>OCFL Object</a> that may exist either as a plain directory, or in the
-            case of a zipped OCFL Object, as the top directory in a zip file. The Object Root is identified by
-            the presence of a NamAsTe file, "0=ocfl_object_1.0".
+            The base directory of an <a>OCFL Object</a>, identified by a NamAsTe file "0=ocfl_object_1.0".
         </dd>
         <dt>
             <dfn>OCFL Storage Root</dfn>:
         </dt>
         <dd>
-            A base directory used to store OCFL Objects, identified by a NamAsTe file, "0=ocfl_1.0".
+            A base directory used to store OCFL Objects, identified by a NamAsTe file "0=ocfl_1.0".
         </dd>
         <dt>
             <dfn>OCFL Version</dfn>:
         <dt>
         <dd>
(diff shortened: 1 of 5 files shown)


draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
-        <dd>The base directory of an <a>OCFL Object</a> that may exist either as a
-          first class directory, or in the case of a zipped OCFL Object, as the top
-          directory in a zip file.</dd>
-        <dt><dfn>OCFL Storage Root</dfn>:</dt>
-        <dd>The base directory of an OCFL storage layout.</dd>
-        <dt><dfn>OCFL Version</dfn>:<dt>
-        <dd>The state of an <a>OCFL object</a>'s content which is constructed using the incremental
-          changes recorded in the sequence of corresponding and prior version directories.</dd>
+        <dd>
+            The file path as it appears on disk including the version directories and the filename, and relative to
```
awoods: The wording here is a bit odd:
```
including the version directories and the filename, and relative to an <a>OCFL Object Root</a>.
```

draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
+            an <a>OCFL Object Root</a>. Existing file paths are used in the <a>Manifest</a> within an
+            <a>Inventory</a>.
+        </dd>
+        <dt>
+            <dfn>Digest</dfn>:
+        </dt>
+        <dd>
+            An algorithmic characterisation of the contents of a file conforming to a standard digest algorithm.
```
awoods: U.S. spelling `characterization`

draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
+        </dt>
+        <dd>
+            A directory for storing information about the content (e.g., actions performed) that may not
+            necessarily be part of the content itself.
+        </dd>
+        <dt>
+            <dfn>Manifest</dfn>
+        </dt>
+        <dd>
+            A complete listing of all files, with their digests, within an OCFL Object and stored in the
```
awoods: It would read better for me if the `and stored in the <a>Inventory</a>` came earlier. Such as:
```
A section of the <a>Inventory</a> containing a complete listing...
```

draft/spec/index.html
rosy1280: The addition of ```version directories``` rather than ```directories``` also ignores the potential for directories within the OCFL object itself.

draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
+            The file path as it appears on disk including the version directories and the filename, and relative to
+            an <a>OCFL Object Root</a>. Existing file paths are used in the <a>Manifest</a> within an
+            <a>Inventory</a>.
+        </dd>
+        <dt>
+            <dfn>Digest</dfn>:
+        </dt>
+        <dd>
+            An algorithmic characterisation of the contents of a file conforming to a standard digest algorithm.
+            Digests are used for recording fixity and for content-addressable storage within an OCFL Object.
```
rosy1280: We were trying to keep definitions to one sentence.  Not an easy thing.  Can we merge these two sentences into one?

draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
-          first class directory, or in the case of a zipped OCFL Object, as the top
-          directory in a zip file.</dd>
-        <dt><dfn>OCFL Storage Root</dfn>:</dt>
-        <dd>The base directory of an OCFL storage layout.</dd>
-        <dt><dfn>OCFL Version</dfn>:<dt>
-        <dd>The state of an <a>OCFL object</a>'s content which is constructed using the incremental
-          changes recorded in the sequence of corresponding and prior version directories.</dd>
+        <dd>
+            The file path as it appears on disk including the version directories and the filename, and relative to
+            an <a>OCFL Object Root</a>. Existing file paths are used in the <a>Manifest</a> within an
```
rosy1280: We were trying to keep definitions to one sentence. Not an easy thing. Can we merge these two sentences into one?

draft/spec/index.html
rosy1280: +1

draft/spec/index.html
rosy1280: The addition of ```version directories``` rather than ```directories``` also ignores the potential for directories within the OCFL object itself.

draft/spec/index.html
rosy1280: Again two sentences...

draft/spec/index.html
```diff
@@ -191,36 +191,80 @@ <h2>Need</h2>
...
+            URI. The object may contain a sequence of versions of the files that represent the evolution of the
+            object's contents.
+        </dd>
+        <dt>
+            <dfn>OCFL Object Root</dfn>:
+        <dt>
+        <dd>
+            The base directory of an <a>OCFL Object</a> that may exist either as a plain directory, or in the
+            case of a zipped OCFL Object, as the top directory in a zip file. The Object Root is identified by
+            the presence of a NamAsTe file, "0=ocfl_object_1.0".
```
rosy1280: Again two sentences....

zimeon:
@rosy1280 I think the key is to keep the definitions short, _usually_ one sentence. However, if one sentence really doesn't work then I would not force awkward language or forgo clarity to follow that rule

draft/spec/index.html
zimeon: IMO the second sentence isn't required here, it is repeating part of the spec

draft/spec/index.html
zimeon: I think the second sentence is actually slightly misleading now... we don't really use **content-addressable storage** in the usual sense (the files most likely have their original names), instead we have a kind of **content reference** in the inventory. My sense is that the two sentences are short so I'd be fine to keep them, but would change "content-addressable storage" to "content references"

ahankinson:
Thanks for the comments -- I've re-worked the definitions according to feedback, with one exception: I kept the reference to the NamAsTe file in OCFL Object Root since it a) mirrors that of the one in OCFL Storage Root, and b) an OCFL Object Root can be identified as such by its presence, so it forms part of the identity (which seems important to note in the definition)

draft/spec/index.html
```diff
@@ -203,40 +203,84 @@ <h2>Need</h2>
...
+            A group of one or more content files and administrative 
+            information, that are together identified by a URI. The object may 
+            contain a sequence of versions of the files that represent the 
+            evolution of the object's contents.
+        </dd>
+        <dt>
+            <dfn>OCFL Object Root</dfn>:
+        <dt>
+        <dd>
+            The base directory of an <a>OCFL Object</a>, identified by a NamAsTe file "0=ocfl_object_1.0".
```
zimeon: [[NAMSATE]] not NamAsTe? Good to link and the spec doesn't use the CamelCase version (though the do explain as "NAMe AS TExt")

draft/spec/index.html
```diff
@@ -203,40 +203,84 @@ <h2>Need</h2>
...
+        </dd>
+        <dt>
+            <dfn>OCFL Storage Root</dfn>:
+        </dt>
+        <dd>
+            A base directory used to store OCFL Objects, identified by a NamAsTe file "0=ocfl_1.0".
```
zimeon: similarly: [[NAMASTE]]

draft/spec/index.html
```diff
@@ -203,40 +203,84 @@ <h2>Need</h2>
...
-        <dd>The base directory of an <a>OCFL Object</a>.</dd>
-
-        <dt><dfn>OCFL Storage Root</dfn>:</dt>
-        <dd>The base directory of an OCFL storage layout.</dd>
-
-        <dt><dfn>OCFL Version</dfn>:<dt>
-        <dd>The state of an <a>OCFL object</a>'s content which is constructed using the incremental changes recorded in
-          the sequence of corresponding and prior version directories.</dd>
+        <dd>
+            The file path of a file on disk, relative to the <a>OCFL Object
```
zimeon: ... of a file on disk **or in an object store**, ...

draft/spec/index.html
```diff
@@ -203,40 +203,84 @@ <h2>Need</h2>
...
+        </dt>
+        <dd>
+            An algorithmic characterization of the contents of a file 
+            conforming to a standard digest algorithm.
+        </dd>
+        <dt>
+            <dfn>Inventory</dfn>:
+        </dt>
+        <dd>
+            A file, expressed in JSON-LD, that tracks the history and current
```
rosy1280: No JSON-LD

zimeon: Well spotted!

ahankinson:
🎉

PR #178, opened 2018-10-01
WIP DNM JSON-LD context

### LINKED ISSUE #41: Develop JSON-LD context specification for OCFL
As per https://github.com/OCFL/spec/issues/1#issuecomment-406052044, point 3.

zimeon:
Closes #41

### DIFF
--- a/draft/context.jsonld
+++ b/draft/context.jsonld
@@ -1,7 +1,46 @@
 {
   "@context": {
     "@version": 1.1,
+    "ocfl": "https://ocfl.io/draft/spec#",
+    "dcterms": "http://purl.org/dc/terms/",
     "id": "@id",
-    "type": "@type"
+    "type": "@type",
+    "digestAlgorithm": {
+      "@id": "ocfl:digestAlgorithm"
+    },
+    "head": {
+      "@id": "ocfl:head"
+    },
+    "manifest": {
+      "@id": "ocfl:manifest"
+    },
+    "versions": {
+      "@id": "ocfl:versions",
+      "@container": "@list",
+      "@type": "ocfl:Version"
+    },
+    "Version": "ocfl:Version",
+    "version": {
+      "@id": "ocfl:version"
+    },
+    "created": {
+      "@id": "dcterms:created",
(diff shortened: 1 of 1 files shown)


draft/context.jsonld
```diff
@@ -1,7 +1,51 @@
...
+      "@id": "ocfl:message"
+    },
+    "user": {
+      "@id": "ocfl:user"
+    },
+    "name": {
+      "@id": "ocfl:name"
+    },
+    "address": {
+      "@id": "ocfl:address"
```
awoods: `address`: are we leaving this open to any kind of address? email, mailing, etc?

draft/context.jsonld
```diff
@@ -1,7 +1,51 @@
...
+    },
+
+    "version": {
+      "@id": "ocfl:version"
+    },
+    "created": {
+      "@id": "dcterms:created",
+      "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+    },
+    "message": {
```
awoods: `message`: Is this intended as a "comment" on OCFL:Objects?

ahankinson: I think it's a comment on a version, like a commit message.

draft/context.jsonld
ahankinson: I think it's safe to restrict it to electronic addresses, but whether that includes e-mail, twitter handles, etc. should be discussed.

ahankinson:
Is this ready for a merge review, or is it still WIP?

zimeon:
Still very much WIP. It is currently incomplete and at the moment I don't think it is possible to have a JSON-LD context that handles our `fixity` section sensibly...

draft/context.jsonld
zimeon: Yeah, am following the spec here which seems OK for now. I think we should think a bit harder about the metadata elements after we have this draft done, they seem rather arbitrary to me at the moment and I'm really not sure that they should all be MUST -- it would be perfectly reasonable to decide to store such metadata within the versions according to local convention, it is not necessary for handling the objects per se

zimeon:
Closing for now, we have a bigger discussion to resolve on #41

PR #194, opened 2018-10-03
Remove JSON-LD from spec, use only JSON

zimeon:
Closes#190


zimeon:
@awoods - have created #195 for impl notes

PR #205, opened 2018-10-09
Fixed: Add a BagIt example

### LINKED ISSUE #189: BagIt example
once @zimeon updates JSON, create an example object using BagIt

ahankinson:
This PR adds a BagIt example.

Fixes #189

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -966,7 +966,7 @@
     <section id="example-bagit-in-ocfl">
         <h2>BagIt in an OCFL Object</h2>
         <p>
-            [[!BagIt]] is a popular file packaging specification, but unlike OCFL it does not provide a mechanism for
+            [[BagIt]] is a common file packaging specification, but unlike OCFL it does not provide a mechanism for
             content versioning. Using OCFL it is possible to store a BagIt structure with content versioning, such
             that when the object state is resolved, it creates a valid BagIt 'bag'. This example will illustrate how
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -953,6 +965,127 @@ <h2>Versioned OCFL Object</h2>
...
+        └── myfirstbag
+            ├── data
+            │   └── 27614-h
+            │       └── images
+            │           ├── q173.png
+            │           └── q173.txt
+            └── manifest-md5.txt
+</pre>
+        <p>
+            The state of the object at version 2 would resemble the following BagIt object:
```
zimeon: "resemble" sounds to weak to me. the state really **is** that Bagit object. Perhaps "be"  or "match"?

draft/spec/index.html
```diff
@@ -953,6 +965,127 @@ <h2>Versioned OCFL Object</h2>
 
     <section id="example-bagit-in-ocfl">
         <h2>BagIt in an OCFL Object</h2>
+        <p>
+            [[!BagIt]] is a popular file packaging specification, but unlike OCFL it does not provide a mechanism for
```
awoods: I believe this should be `[[BagIt]]` instead of the normative `[[!BagIt]]`.

awoods: I would advocate for less subjective language such as "popular". Maybe replace "popular" with "common"?

ahankinson: Oh, I thought it was the other way around. I'll fix it.

ahankinson: Awesome? Tubular? Totally Rad? 

OK.

PR #207, opened 2018-10-09
Fixed: Attempt to make the administrative info block less confusing

### LINKED ISSUE #204: Confusing paragraph
Section 3 preamble:

> The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to store their own metadata serializations as files in an OCFL Object, but, in the context of the OCFL Specification these are considered to be part of the contents of an OCFL Object.

ahankinson:
Fixes #204

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -293,9 +293,8 @@
         specification.
     </p>
     <p>
-        An OCFL Object also has administrative information that identifies a folder as an OCFL Object, and also
-        provides a means of tracking changes to the contents of the object over time. Administrative files, such as
-        the <a>Inventory</a> files, are not considered to be part of the content of the object.
+        An OCFL Object includes administrative information that identifies a directory as an OCFL Object, and also
+        provides a means of tracking changes to the contents of the object over time.
     </p>
     <p>
         An OCFL Object is therefore:
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -281,9 +281,9 @@
         specification.
     </p>
     <p>
-        The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -281,9 +281,9 @@ <h2>OCFL Object</h2>
         specification.
     </p>
     <p>
-        The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to
-        store their own metadata serializations as files in an OCFL Object, but, in the context of the OCFL
-        Specification these are considered to be part of the contents of an OCFL Object.
+        An OCFL Object also has administrative information that identifies a folder as an OCFL Object, and also
+        provides a means of tracking changes to the contents of the object over time. Administrative files, such as
+        the <a>Inventory</a> files, are not considered to be part of the content of the object.
```
awoods: Maybe it belongs in the "Implementation Notes", but this updated text loses the recommendation to consider client-provided metadata as "content".

draft/spec/index.html
```diff
@@ -281,9 +281,9 @@ <h2>OCFL Object</h2>
         specification.
     </p>
     <p>
-        The administrative information is specific to OCFL Objects. Implementers of OCFL are strongly encouraged to
-        store their own metadata serializations as files in an OCFL Object, but, in the context of the OCFL
-        Specification these are considered to be part of the contents of an OCFL Object.
+        An OCFL Object also has administrative information that identifies a folder as an OCFL Object, and also
```
awoods: Suggestion: 
```
An OCFL Object [also has > includes] administrative information that identifies a [folder > directory] as an OCFL Object
```

draft/spec/index.html
zimeon: I think here it is OK to lose the comment about content including metadata here. It is in the bullet 1 just below

PR #214, opened 2018-10-10
Fixed: Add text about storage extensions

### LINKED ISSUE #159: Flesh out Storage Root Extension section
currently blank

ahankinson:
Fixes #159

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -798,9 +798,9 @@
     <section id="root-extensions">
         <h2>Extensions</h2>
         <p>
-            The behavior of the storage root may be extended to support features from other specifications.  An OCFL
-            validator MUST ignore any files in the storage root it does not understand, but these files MUST not appear
-            outside of the storage root.
+            The behavior of the storage root may be extended to support features from other specifications.  
+            An OCFL validator MUST ignore any files in the storage root it does not understand. Additional 
+	    files MUST NOT appear in other directories under the storage root.
         </p>
         <blockquote class="informative">
             <p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -798,9 +798,9 @@
     <section id="root-extensions">
         <h2>Extensions</h2>
         <p>
-            The behavior of the storage root may be extended to support features from other specifications.  
+            The behavior of the storage root may be extended to support features from other specifications.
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -797,6 +797,18 @@ <h2>Storage Hierarchies</h2>
 
     <section id="root-extensions">
         <h2>Extensions</h2>
+        <p>
+            The behavior of the storage root may be extended to support features from other specifications.  An OCFL
+            validator MUST ignore any files in the storage root it does not understand, but these files MUST not appear
```
zimeon: "these files" seems wrong as we are not talking about ones in the storage root. Also s/not/NOT/ in "MUST not".  How about:

> An OCFL validator MUST ignore any files in the storage root it does not understand. Additional files MUST NOT appear in other directories under the storage root.

draft/spec/index.html
```diff
@@ -797,6 +797,18 @@ <h2>Storage Hierarchies</h2>
 
     <section id="root-extensions">
         <h2>Extensions</h2>
+        <p>
+            The behavior of the storage root may be extended to support features from other specifications.  
+            An OCFL validator MUST ignore any files in the storage root it does not understand. Additional 
+	    files MUST NOT appear in other directories under the storage root.
```
awoods: Minor: indent is misaligned.

awoods:
We have an opportunity to provide guidance on the approach to Storage Root extension.
@ahankinson : are you suggesting we stop our guidance at:
> Extension can happen however you want, as long as:
> 1. It is noted in one or more files in the Storage Root, and
> 1. It is not noted within a directory off of the Storage Root

?

ahankinson:
Yes? I'm not sure what else we can say... suggestions?

awoods:
@ahankinson : we have discussed approaches such as Namaste indicators... but I am ok with the less specific approach in this PR.

ahankinson:
I'm pretty sure that what we settled on in the last editors call is that we would not try to define any NamAsTe indicators, as they are out-of-scope for the OCFL work, but that we could allow them if desired through the extensions mechanism in order to support other specifications.

zimeon:
I agree that we decided not to define any indicators or even a mechanism to do so (though I'm still interested in that) at this stage. Will merge so that this is ready for call.

Issue #218, opened 2018-10-10
Clarify the ordering or non-ordering of versions


dbrower:
Is the version list in the inventory file ordered or unordered?

PR #220, opened 2018-10-11
Clarify non-ordered nature of version items in inventory

awoods:
Resolves: https://github.com/OCFL/spec/issues/218

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -539,7 +539,8 @@
             <h2>Versions</h2>
             <p>
                 An OCFL Object Inventory MUST include a block for storing versions. This block MUST have the
-                key of <code>versions</code> within the inventory, and it MUST be an array.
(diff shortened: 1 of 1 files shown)


ahankinson:
It would probably be helpful to explain why ( version directories are already ordered) since a JSON array is an ordered data structure.

zimeon:
This issue does beg the question of whether there should be a hash of versions indexed by the version directory:

```
"head": "v2",      
"versions": {
  "v1": { ...stuff for v1... },
  "v2": { ...stuff for v2...}
}
```

did we talk about that?

ahankinson:
No, we didn't talk about that. I would be fine with this solution as well. It would also make it easy to say that the value of `head` must be a key in the versions object.

zimeon:
I'm going to bounce the question above up to an issue

PR #227, opened 2018-10-12
WIP: Filesystem Interoperability Notes

ahankinson:
A first pass at filesystem interoperability considerations.

Open for comment and review, not for merging (yet)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -340,6 +340,35 @@
         ├── inventory.json
         └── inventory.json.sha512
         </pre>
+
+        <section id="object-filesystem-restrictions">
+            <h2>Filesystem Restrictions</h2>
+            <p>
+                Within an OCFL Object and its associated <a>Inventory</a>, the following restrictions are applied
+                to filesystem features. These restrictions are provided in an attempt to normalize file handling
+                behaviors and addressability of files within the Inventory.
+            </p>
+            <ul>
+                <li>
+                    <strong>Symbolic links</strong> (aka 'shortcuts' or 'aliases') MUST not be used within an
+                    OCFL Object. These are not portable and may lead to undefined behaviours in client applications
+                </li>
+                <li><strong>File paths</strong> MUST use the "forward-slash" ('/') as a path separator. Some
+                    filesystem use other characters such as the "back-slash" ('\') or "colon" (':') as a
+                    path delimiter. These MUST be normalized to the forward slash.
+                </li>
+                <li>
+                    <strong>File permissions</strong> MAY be applied to files in an OCFL Object; however, implementers
+                    should be aware that these are not portable across filesystems, and so should not be relied on
+                    to protect sensitive content.
+                </li>
+                <li>
+                    The <strong>character encoding</strong> of the filesystem and Inventory SHOULD be
+                    Unicode-compatible, either UTF-8, UTF-16, or UCS-2. Implementers may experience problems
+                    integrating files from filesystems that use other encodings.
+                </li>
+            </ul>
+        </section>
     </section>
 
     <section id="object-conformance-declaration">
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -348,24 +348,35 @@
                 to filesystem features. These restrictions are provided in an attempt to normalize file handling
                 behaviors and addressability of files within the Inventory.
             </p>
+            <p>
+                If use of these features are required to preserve an exact copy of the content (for example,
+                preserving a hard-drive image), implementers SHOULD consider using a packaging format,
+                such as ZIP or TAR files or ISO disk images, to preserve the content.
+            </p>
             <ul>
                 <li>
                     <strong>Symbolic links</strong> (aka 'shortcuts' or 'aliases') MUST NOT be used within an
                     OCFL Object. These are not portable and may lead to undefined behaviours in client applications
                 </li>
-                <li><strong>File paths</strong> MUST use the "forward-slash" ('/') as a path separator. Some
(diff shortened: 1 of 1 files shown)


ahankinson:
Any thoughts?

draft/spec/index.html
```diff
@@ -340,6 +340,47 @@ <h2>Object Structure</h2>
...
+                    to protect sensitive content. This is also the case for extended filesystem flags, such as
+                    Access Control Lists or Hidden files.
+                </li>
+                <li>
+                    The <strong>character encoding</strong> of the filesystem and Inventory SHOULD be
+                    Unicode-compatible, either UTF-8, UTF-16, or UCS-2. Implementers may experience problems
+                    integrating files from filesystems that use other encodings.
+                </li>
+                <li>
+                    Some filesystems are not <strong>case sensitive</strong>, meaning two file names that differ only
```
awoods: Can we add a normative `MUST`, `SHOULD`, or `MAY` in this last item?

draft/spec/index.html
```diff
@@ -340,6 +340,47 @@ <h2>Object Structure</h2>
...
+                    or "colon" (':') as a path delimiter.
+                </li>
+                <li>
+                    <strong>File permissions</strong> MAY be applied to files in an OCFL Object; however, implementers
+                    should be aware that these are not portable across filesystems, and so should not be relied on
+                    to protect sensitive content. This is also the case for extended filesystem flags, such as
+                    Access Control Lists or Hidden files.
+                </li>
+                <li>
+                    The <strong>character encoding</strong> of the filesystem and Inventory SHOULD be
```
zimeon: I don't know what it means to talk about character encodings of a **filesytem** -- I think our expectation here is bytestream fidelity, without that all is lost! I would thus rephrase this paragraph to talk only about inventory files

draft/spec/index.html
```diff
@@ -340,6 +340,47 @@ <h2>Object Structure</h2>
...
+                    Unicode-compatible, either UTF-8, UTF-16, or UCS-2. Implementers may experience problems
```
zimeon: JSON itself is defined over UTF-8, UTF-16 or UTF-32 (with either by order for UTF-16 or UTF-32), see https://tools.ietf.org/html/rfc4627#section-3 . I do not know whether all JSON parsers have good support for all of these encodings. There is no note about UCS-2 encoding so I'm not sure how we imagine that to be handled? I feel a little out of my comfort zone with this question but I think we need to be more explicit and tie to the JSON spec. The current  text seems to raise more questions than it answers in my mind.

draft/spec/index.html
```diff
@@ -340,6 +340,47 @@ <h2>Object Structure</h2>
...
+                <li>
+                    <strong>Symbolic links</strong> (aka 'shortcuts' or 'aliases') MUST NOT be used within an
+                    OCFL Object. These are not portable and may lead to undefined behaviours in client applications
+                </li>
+                <li>References to <strong>file paths</strong> within the <a>inventory</a> MUST use the "forward-slash"
+                    ('/') as a path separator. Some filesystem use other characters such as the "back-slash" ('\')
+                    or "colon" (':') as a path delimiter.
+                </li>
+                <li>
+                    <strong>File permissions</strong> MAY be applied to files in an OCFL Object; however, implementers
```
zimeon: File permissions are unavoidably applied to any file in all current filesystems I know about so the MAY here seems odd. I also don't think we should introduce fuzzy terms like ACLs and hidden files. I think we should make a simpler statement along the lines of:

> File permissions are not portable across filesystems and are not expected to be preserved by OCFL clients.

draft/spec/index.html
zimeon: This seems like a significant question. We either allow a degree of incompatibility or we are rather restrictive. Neither is very appealing. Bagit takes the first approach but gives a description of the issue: https://tools.ietf.org/html/draft-kunze-bagit-17#section-6.1.1.1
so I think I would tend toward that approach. A link to BagIt might be nice here but essentially this is the tack @ahankinson has taken in the PR

zimeon:
I wonder whether some of these questions should be elevated to issues for discussion?

neilsjefferies:
Windows is a real pain here. Under the hood, NTFS allows / and \ to be interchangeable directory separators, and it is also case sensitive and Unicode supporting. However, many of its user space tools are differ since they also support FAT variants with their variable handling of these aspects.

draft/spec/index.html
srerickson: Hi, I'm new here but wanted to add that it may be helpful to be more specific about Unicode Normalization, particularly given the difficulties encountered by folks working on BagIt. (See [here](https://github.com/LibraryOfCongress/bagit-python/pull/82)).

draft/spec/index.html
srerickson: @rosy1280 I think the edge case to consider is when a repository is rebuilt on a file system that handles filenames differently than the filesystem the OCFL Object was created on. In that situation (as in the example with BagIt) it might be possible for the filenames in the inventory to differ from the actual filenames (even though they both look the same, visually).

draft/spec/index.html
srerickson: @rosy1280 I think the edge case to consider is when a repository is rebuilt on a file system that handles filenames differently than the filesystem the OCFL Object was created on. In that situation (as in the example with BagIt) it might be possible for the filenames in the inventory to differ from the actual filenames (even though they both look the same, visually).

ahankinson:
Editors call: 05/12/2018 After discussion we'll close this and break specific discussions out to other tickets / PRs.

PR #225, opened 2018-10-12
Fixed: Replace logical with existing

ahankinson:
Fixes #221

Issue #221, opened 2018-10-11
Section 3.5.2 "Manifest" to reference existing file paths


awoods:
Both mentions of "logical file paths" in this section should be changed to "existing file paths".

PR #228, opened 2018-10-12
Fixed: Specify that EFPs must be relative to OR.

ahankinson:
This PR specifies that existing file paths MUST be given relative to the Object Root.

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -513,11 +513,8 @@
                 The value of the <code>manifest</code> key is a JSON object, with keys corresponding to the digests
                 of every content file in all versions of the <a>OCFL Object</a>. The value for each key is an array
                 containing the <a>existing file path</a>s of files in the OCFL Object that have content with the
-                given digest.
-            </p>
-            <p>
-                <a data-lt="existing file path">Existing file paths</a> within a manifest block MUST be given relative
-                to the <a>OCFL Object Root</a>.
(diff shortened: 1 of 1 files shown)


ahankinson:
It is mentioned in the terminology section, but I think we decided that normative language is not used in terminology. So this is primarily to apply the normative language to the requirement for EFPs.

zimeon:
I agree with @ahankinson that it is good to make this clear statement in the main text, even if it duplicates the definition. I guess this will have conflicts given other LFP->EFP changes. I suggest that the sentence should be part of the previous paragraph rather than living on its own.

awoods:
The extra clarity works for me.

draft/spec/index.html
```diff
@@ -515,6 +515,10 @@ <h2>Manifest</h2>
                 containing the <a>logical file path</a>s of files in the OCFL Object that have content with the
                 given digest.
             </p>
+            <p>
```
zimeon: Suggest combine into preceding para but otherwise good by me

ahankinson:
The changes here will conflict with changes from fixing #221, in PR #225. I've rolled the changes into that PR, so I will close this one.

ahankinson:
*sigh* ignore me. Re-opening b/c that one was already merged.

ahankinson:
@awoods @zimeon could you merge this please?

zimeon:
Editorial change, merging on 3...

PR #233, opened 2018-10-16
Fixed: Change to a JSON Object for versions

### LINKED ISSUE #231: Should we have a versions object indexed by version directory?
At present we have a `versions` array that has an object for each version. This leads to questions about the ordering of the items in that array (cf. #218, #220).  Should we instead have a `versions` object that is keyed by the version directory, with the value of `head` providing a pointer in to the key of the latest version:

```
{
  ...
  "head": "v2",      
  "versions": {
    "v1": { ...stuff for v1... },
    "v2": { ...stuff for v2... }
  }
}
```

ahankinson:
Fixes #231

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -539,12 +539,9 @@
             <h2>Versions</h2>
             <p>
                 An OCFL Object Inventory MUST include a block for storing versions. This block MUST have the
-                key of <code>versions</code> within the inventory, and it MUST be an array. There is no requirement
-                that the ordering of versions in this array represent a chronological or order of any kind.
-            </p>
-            <p>
-                The <code>versions</code> array MUST be a list of one or more JSON objects, one for each
-                <a href="#version"></a>.
+                key of <code>versions</code> within the inventory, and it MUST be a JSON object. The keys of this object
+                MUST correspond to the names of the <a href="#version-directories">version directories</a> used. The
+                values MUST be another JSON object that describes this version.
             </p>
             <section id="version">
                 <h2>Version</h2>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -539,12 +539,9 @@ <h2>Manifest</h2>
             <h2>Versions</h2>
             <p>
                 An OCFL Object Inventory MUST include a block for storing versions. This block MUST have the
-                key of <code>versions</code> within the inventory, and it MUST be an array. There is no requirement
-                that the ordering of versions in this array represent a chronological or order of any kind.
```
awoods: This update drops the [clarification of ordering](https://github.com/OCFL/spec/issues/218).

ahankinson: As JSON objects are explicitly unordered, I don't know that we need that?

>  An object is an unordered collection of zero or more name/value
>  pairs, where a name is a string and a value is a string, number,
>  boolean, null, object, or array.

https://tools.ietf.org/html/rfc7159

zimeon: I wondered whether we need any clarification of ordering if we have an object with version keys. I just took a look at the [JSON spec](https://tools.ietf.org/html/rfc7159#section-4) and it leaves the door open for the significance of order.  So, I think that means we should keep this sentence (modulo s/array/object/) but perhaps move it to the top of the [inventory section](https://ocfl.io/draft/spec/#inventory) as a general statement that applies to all objects.

zimeon: The crack in the door I see in RFC7159 is:

> JSON parsing libraries have been observed to differ as to whether or
   not they make the ordering of object members visible to calling
   software.

zimeon: Maybe second sentence in 3.5 intro which is expanded to cover the arrays we use (in which order is also not significant):

> The order of entries in both the [[JSON]] objects and arrays used in inventory files has no significance.

rosy1280: I agree with adding a note to section 3.5 that @zimeon suggests two comments up.

rosy1280:
we assumed that @neilsjefferies comment above was approval that satisfies @zimeon 's request

PR #236, opened 2018-10-17
Fixed: Add content directory

### LINKED ISSUE #230: What if my content includes files called `inventory.json` or `inventory.json.sha512`?
I see two options:

1) With the spec as currently written there would need to be a note explaining that these are special cases where the existing file path cannot be the original filepath, it must be renamed inside the OCFL object, e.g.:

```
[object root]
|- v1
|    |- inventory.json   <-- the actual inventory, not the file so named in v1 content
|    |- inventory.json.sha512   <-- the actual digest sidecar
|    |- inventory.json_moved   <-- the content called inventory.json in v1 state
|    \- inventory.json.sha512_moved  <-- the content called inventory.json.sha512 in v1 state
\...
```

2) Use another directory as is done in [BagIt](https://tools.ietf.org/html/draft-kunze-bagit-17#section-2) (where it is called `data`) so that there is clean separation:

```
[object root]
|- v1
|    |- inventory.json   <-- the actual inventory, not the file so named in v1 content
|    |- inventory.json.sha512   <-- the actual digest sidecar
|    \- data
|          |- inventory.json  <-- the content called inventory.json in v1 state
|          \- inventory.json.sha512  <-- the content called inventory.json.sha512 in v1 state
\...
```

ahankinson:
Fixes #230

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -331,17 +331,15 @@
             The basic OCFL Object structure has a minimal set of files and directories necessary to support content
             storage and object validation. The minimum required is shown in the following figure:
         </p>
-        <pre>
+<pre>
 [object_root]
     ├── 0=ocfl_object_1.0
     ├── inventory.json
     ├── inventory.json.sha512
     └── v1
         ├── inventory.json
-        ├── inventory.json.sha512
-        └── content
-            └── .keep
-        </pre>
+        └── inventory.json.sha512
+</pre>
     </section>
 
     <section id="object-conformance-declaration">
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -385,7 +385,7 @@
         </p>
         <p>
             Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
-            SHOULD NOT contain a <code>content</code> subdirectory otherwise. There MUST be no other files or
+            SHOULD NOT contain a <code>content</code> sub-directory otherwise. There MUST be no other files or
             directories as children of a version directory, other than an <a href="#inventory">inventory file</a>,
             an <a href="#inventory-digest">inventory digest</a>, and a <code>content</code> directory.
         </p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -338,7 +338,9 @@
     ├── inventory.json.sha512
     └── v1
         ├── inventory.json
-        └── inventory.json.sha512
+        ├── inventory.json.sha512
+        └── content
+            └── .keep
         </pre>
     </section>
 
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -383,6 +385,12 @@ <h2>Version Directories</h2>
             Empty directories within a version directory are not permitted. Otherwise empty directories MAY be
             maintained by creating a <code>.keep</code> file within that directory.
         </p>
+        <p>
+            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
```
awoods: Text looks good.

draft/spec/index.html
```diff
@@ -383,6 +385,12 @@ <h2>Version Directories</h2>
...
+            SHOULD NOT contain a <code>content</code> subdirectory otherwise. There MUST be no other files or
```
awoods: The other instances of "subdirectory" are spelled with a hyphen: "sub-directory".

draft/spec/index.html
```diff
@@ -876,21 +885,26 @@ <h2>Versioned OCFL Object</h2>
...
-    │   └── inventory.json.sha512
+    │   ├── inventory.json.sha512
+    │   └── content
+    │       ├── foo
+    │       └── bar.xml
     └── v3
         ├── inventory.json
-        └── inventory.json.sha512
+        ├── inventory.json.sha512
+        └── content
```
awoods: We are suggesting not to include empty `content` directories.

draft/spec/index.html
ahankinson: Thanks; it's really the only substantive change; the rest is just fixing the examples.

draft/spec/index.html
ahankinson: fixed

draft/spec/index.html
ahankinson: fixed

draft/spec/index.html
```diff
@@ -338,7 +338,9 @@ <h2>Object Structure</h2>
     ├── inventory.json.sha512
     └── v1
         ├── inventory.json
-        └── inventory.json.sha512
+        ├── inventory.json.sha512
+        └── content
+            └── .keep
```
zimeon: I do not think we should show a `content` dir with a `.keep`.  Technically, the `content` and inventory files in the version should not be present anyway because the current description says "The **minimum required** is." -- neither are required for the minimum. However, I think it would be better to instead say something like:

> The OCFL Object structure organizes content files and administrative information in order to support content storage and object validation. The structure for an object with one version is shown in the following figure:
>
> ```
> [object_root]
>    ├── 0=ocfl_object_1.0
>    ├── inventory.json
>    ├── inventory.json.sha512
>    └── v1
>        ├── inventory.json
>        ├── inventory.json.sha512
>        └── content
>            └── ... content files ...

zimeon: Somehow this got marked outdated in a race condition --- I suggest also that we must reword the intro text because this isn't the minimal form

ahankinson: WIll do.

zimeon:
I suggest also that we must reword the intro text because the example isn't the minimal form. Technically, neither the `content` nor inventory files in the version should be present for minimal. However, I think it would be better to instead say something like:

> The OCFL Object structure organizes content files and administrative information in order to support content storage and object validation. The structure for an object with one version is shown in the following figure:
>
> ```
> [object_root]
>    ├── 0=ocfl_object_1.0
>    ├── inventory.json
>    ├── inventory.json.sha512
>    └── v1
>        ├── inventory.json
>        ├── inventory.json.sha512
>        └── content
>            └── ... content files ...

ahankinson:
Done.

ahankinson:
🙌 🎉 :shipit:

Issue #208, opened 2018-10-09
Do we really need the objects under a storage root to have the same OCFL version as the storage root, or as each other


zimeon:
Currently, the [root structure](https://ocfl.io/draft/spec/#root-structure) section says:

> OCFL Objects within a given OCFL Storage Root MUST be the same as the OCFL specification version as declared in the Root Conformance Declaration.

and reiterates as

> ... For example, an institution may have a "newspapers" project that standardizes on OCFL 1.0, but then may have a later "letters" project that wishes to use OCFL 1.2. These two projects would exist within two separate OCFL Storage Roots.

I'm not sure that it is necessary for either:
  1. the storage root version to match the version of the objects
  2. (requiring 1 but also) the objects to all have the same version, that seems to impose a big migration burden

julianmorley:
I agree. I'd rather the root conformance declaration be seen as a statement that 'in this place, you may find OCFL objects, and the highest version of those objects may be X', rather than 'this place may only contain OCFL objects of version X and nothing else'.

rosy1280:
I would like to suggest that:

1. Storage root declaration conformity means that the storage root conforms to section 4 (i.e. the Storage Root section) of the specification; and
1. OCFL Objects may conform to lower versions of the specification as identified in the objects conformance declaration

PR #237, opened 2018-10-17
Remove example "mis-"requiring multiple storage roots

awoods:
Follow-on to: https://github.com/OCFL/spec/issues/208


ahankinson:
Speedy mode on 3. Zoom zoom.

Issue #238, opened 2018-10-17
Release Alpha 0.1


awoods:
* [x] get alpha spec up (#239)
  * [x] fix schema (#241)
  * [x] add links on homepage (#244)
  * [x] make announcements

zimeon:
#239 still needs approval from @julianmorley and @neilsjefferies

zimeon:
yup, I updated the checklist at top

neilsjefferies:
OK

awoods:
Resolved: https://groups.google.com/d/msg/ocfl-community/BB6anEmITEg/icN1mMe-BAAJ

PR #241, opened 2018-10-17
JSON schema for 0.1

zimeon:
Update schema for the new inventory structure as part of #238 0.1 alpha release

See https://github.com/zimeon/ocfl-py/blob/master/README_inventory_validation.md for a way to build an example inventory and to run a test against the schema.

### DIFF
--- a/draft/spec/inventory_schema.json
+++ b/draft/spec/inventory_schema.json
@@ -1,5 +1,5 @@
 {
-    "$id": "https://ocfl.io/draft/spec/inventory_schema.json",
+    "$id": "https://ocfl.io/0.1/spec/inventory_schema.json",
     "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "JSON Schema for OCFL Inventory files",
     "description": "Note that this schema does not permit extension keys, only keys defined in the OCFL Specification are admitted.",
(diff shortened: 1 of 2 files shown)


draft/spec/inventory_schema.json
```diff
@@ -34,44 +34,44 @@
...
-                                "type": "string"
-                            },
-                            "name": {
-                                "type": "string"
-                            }
+            "description": "Each key is a version directory name with version objects as values",
+            "type" : "object",
+            "additionalProperties": false,
+            "patternProperties": {
+                "^v\\d+$": {
```
ahankinson: Can we use this regex to forbid non-zero initial numbers?

zimeon: No, that would be wrong. v99999 is allowed if you start with v1

zimeon: ...and v09999 is right if you start with v00001

ahankinson: Hm. So how will we validate:

"If zero-padded version directory numbers are used then they must start with the prefix v and then a zero. For example, in an implementation that uses five digits for version directory names then v00001 to v09999 are allowed, v10000 is not allowed."

zimeon: With code rather than JSON schema. There are lots of parts of the spec than cannot be validated with the schema (like the checksum actually being right). There may be some more complex JSON schema foo that would get us a bit further, but it certainly won't get us all the way

ahankinson:
Since this is a non-normative change I will merge it (as discussed on the editors call).

PR #243, opened 2018-10-18
Added notes on Filesystem Features

neilsjefferies:
First stab at #212

Issue #212, opened 2018-10-10
Normative opinions on filesystem features


ahankinson:
I'm wondering if there is a place somewhere in the spec (i.e., normative) where we can say something about the use of symlinks, and that they MUST not be used to implement versioning.

Reasons:
 1. Not portable
 2. Not implementable on Object Stores
 3. May cause problems or unexpected behaviours with some clients
 4. Inconsistencies with relative v. absolute paths

Are there other filesystem-level features that we should mention? Permissions? Extended attributes? Character sets? Path length restrictions / considerations? (Our storage guys want the shortest path possible for performance reasons) Restricted words / characters?

neilsjefferies:
https://en.wikipedia.org/wiki/Comparison_of_file_systems

neilsjefferies:
I've been thinking about this and maybe the best way would be to say that OCFL treats symlinks and hard links as if the file was in situ for the purposes of operations and validation. In other words, OCFL does not recognise them as anything other than files.

A side effect is that you *could* use symlinks for dedupe and the corresponding OCFL inventory would appear to be an undeduplicated one.- and would/should validate and copy as if it were so.

ahankinson:
I've had problems in the past with file operations that helpfully 'resolve' symlinks, meaning copying from one FS to another can result in vastly different amounts of storage used. Or, the opposite, where paths are given as absolute symlinks, and thus everything breaks.

I think it would be easier all round (validation, checksumming, client operations) if we were to consider symlinks *verboten*.

neilsjefferies:
That means we have to detect symlinks and hard links and that makes any validator rather less portable.  I don't think OCFL should worry about how a filesystem path gets to a bytestream - or else we end up delving into the guts of filesystems. Do we want to detect mount points too, how about loopback and overlay mounts? Hierarchical filesystems?

neilsjefferies:
As per my note - that is not actually very practical. Either we adopt a fairly minimal approach to filesystem operations or we get mired in filesystem specific features.

ahankinson:
It is very practical. It's taking an opinionated stand on a well-understood and widely implemented filesystem feature that can have long-reaching implications on bitstream addressability across data transfers and technology migrations. I think we should come down on the side of 'no', for the reasons stated in my first post.

[...]

PR #246, opened 2018-10-18
Filesystem features brought over from 0.1

neilsjefferies:


awoods:
As a procedural note, even in speedy-mode (which we may or may not want to retain) the process for merging is:
- two reviews are required
- author can never merge their own pull-request

The rationale being that we strive for agreement and consensus before committing changes.

zimeon:
Addresses #212 in place of #243 (which was accidentally against /0.1/ rather than /draft/)

PR #248, opened 2018-10-19
Remove 'v2' key in Simple Object JSON example

tomwrobel:
Previous request was made against what I think is the wrong version of the spec file.


ahankinson:
Merging as a minor update.

Issue #240, opened 2018-10-17
Suggestions from review of alpha draft


zimeon:
### Intro

> \< This specification is a normative specification which describes the nature of an OCFL Object (the "object-at-rest"). 

The "specification ... specification" seems redundant and we don't mention the layout part. I suggest:

> \> This normative specification describes the nature of an OCFL Object (the "object-at-rest") and the arrangement of OCFL Objects under an OCFL Storage Root.  

### Need

change -- to emdash (twice)

"file and directory layout on disk" --> add "or in an object store"?

### 2. Terminology

> \< Logical State: A grouping of logical file paths reflecting the state of the content for a given version.

I think the logical state includes the content at the file paths too, so maybe:

> \< Logical State: A grouping of logical file paths with particular content reflecting the state of the content for a given version.

> \< Logs Directory: A directory for storing information about the content (e.g., actions performed) that may not necessarily be part of the content itself.

"may not necessarily be part of the content itself" --> "is not part of the content itself"

> \> Logs Directory: A directory for storing information about the content (e.g., actions performed) that is not part of the content itself.

### 3.3 Version Directories

> \< Empty directories within a version directory are not permitted. Otherwise empty directories may be maintained by creating a .keep file within that directory.
>  
> \< Version directories must have a sub-directory called `content` if there are files present, and should not contain a content sub-directory otherwise. There must be no other files or directories as children of a version directory, other than an inventory file, an inventory digest, and a `content` directory.

I think these two paragraphs should be switched and the "Empty directories..." one then rewritten as a normative statement. Also rewording to avoid otherwise at the start of the sentence which is easy to misread (or at least start to misread):

> \> There MUST NOT be empty directories within a version's `content` directory. Directories that would otherwise be empty MAY be maintained by creating a `.keep` file within the directory.

but also, is it perhaps a requirement that:

> \> All files within a version's `content` directory MUST be referenced in the Inventory `manifest`.

this could be a lead-in statement to the empty directory paragraph.

### 3.4 Digests

> \< OCFL Objects should use `sha512` by default.

insert "digests":

> \> OCFL Objects should use `sha512` digests by default.

and on through that section in most places a digest algorithm name is mentioned.

### 3.5.1 Basic Structure

> < `id`: A unique identifier for the OCFL Object. This must be unique in the local context, and should be either a URI or URN. There is no expectation that values given as URIs are resolveable as URLs.

[...]

PR #254, opened 2018-10-27
Tweaks to state and logs terminology

zimeon:
Part of #240

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -238,14 +238,15 @@
             <dfn>Logical State</dfn>:
         </dt>
         <dd>
-            A grouping of logical file paths reflecting the state of the content for a given version.
+            A grouping of logical file paths with particular content reflecting the state of the object content for
+            a given version.
         </dd>
         <dt>
             <dfn>Logs Directory</dfn>:
         </dt>
         <dd>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -238,14 +238,15 @@ <h2>Terminology</h2>
             <dfn>Logical State</dfn>:
         </dt>
         <dd>
-            A grouping of logical file paths reflecting the state of the content for a given version.
+            A grouping of logical file paths with particular content reflecting the state of the object content for
```
ahankinson: 'content' ... 'content' is awkward. Not sure what the 'particular content' here refers to?

awoods: The previous text for this sentence seems more clear to me. Is there a particular essence that this update is trying to clarify?

zimeon: Yes, the old sentence is wrong in my opinion because it says the state is only the file paths. The state is actually the pointers to the (particular) content and their paths. I agree that the particular didn't work, proved in part because it didn't make sense to either of you. How about:

> A grouping of logical file paths tied to their corresponding bitstreams that reflect the state of the object content for a given version.

ahankinson:
Merging under the 72h rule

PR #255, opened 2018-10-27
Clarify rules for version directories

zimeon:
Part of #240


ahankinson:
Merging under the 72h rule

PR #256, opened 2018-10-27
Tweak digest wording

zimeon:
Part of #240

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -403,17 +403,14 @@
             through hardware degradation, accident, or malicious actors.
         </p>
         <p>
-            OCFL Objects SHOULD use <code>sha512</code> digests by default. The choice of the <code>sha512</code>
-            digest algorithm recognizes that it has no known collision vulnerabilities and is less computationally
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -403,13 +403,13 @@ <h2>Digests</h2>
             through hardware degradation, accident, or malicious actors.
         </p>
         <p>
-            OCFL Objects SHOULD use <code>sha512</code> by default. The choice of <code>sha512</code> recognizes
-            that it has no known collision vulnerabilities and is less computationally intensive to compute than
-            <code>sha256</code>, [[Stop-Using-SHA-256]].
+            OCFL Objects SHOULD use <code>sha512</code> digests by default. The choice of the <code>sha512</code>
```
awoods: Since the sentence before states:
> Digests play two roles in an OCFL Object.

I wonder if this sentence should begin with:
> For content addressability OCFL Objects SHOULD...

ahankinson: 👍

zimeon: Agreed but I dislike the existing use of the phrase "For content addressability". I'll attempt a larger rework to address both issues...

Issue #219, opened 2018-10-10
Logical file paths - directory separators


bcail:
In the Version objects, should the logical file path values in the state block allow windows or unix directory separators (eg. "dir/file.txt" vs "dir\file.txt")? Should a file name in a logical path be allowed to contain any directory separator characters (eg. "dir/file_with_/_char.txt)? Or should OCFL even care what's in that logical file path?

awoods:
As a point of reference, [BagIt](https://tools.ietf.org/html/draft-kunze-bagit-08) takes the following approach:
   - https://tools.ietf.org/html/draft-kunze-bagit-08#section-7.2.2
> As specified above, only the Unix-based path separator ('/') may be
   used inside filenames listed in BagIt manifests and "fetch.txt"
   files.  When bags are exchanged between Windows and Unix platforms,
   care should be taken to translate the path separator as needed.

ahankinson:
Related to #243

zimeon:
I think we should follow Bagit here and make a normative statement in the introduction to [Inventory](https://ocfl.io/draft/spec/#inventory) that

> The forward slash (`/`) path separator MUST be used in existing and logical file paths in then `manifest`, `fixity` and `state` blocks within the inventory. Implementations that target systems using other separators will need to translate paths appropriately.

PR #263, opened 2018-11-07
Fixed: Clarify Namaste requirements

### LINKED ISSUE #261: NAMASTE files are not supposed to be zero size
We say in https://ocfl.io/draft/spec/#object-conformance-declaration :

> The version declaration MUST be formatted according to the [NAMASTE] specification. It MUST be an empty file in the base directory of the object giving the OCFL Object version in the filename. 

whereas the NAMASTE spec says:

> What's inside the tag file named "D=tvalue" is the full value string from which tvalue is derived.

so the contents would be expected to be `ocfl_object_1.0\n` or similar. NAMASTE does not precisely define the contents but I think we should say that the contents are exactly that. Empty is not following NAMASTE however.

ahankinson:
Fixes #261

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -349,11 +349,12 @@
     <section id="object-conformance-declaration">
         <h2>Object Conformance Declaration</h2>
         <p>
-            The version declaration MUST be formatted according to the [[!NAMASTE]] specification. It MUST be an empty
-            file in the base directory of the object giving the OCFL Object version in the filename. The filename MUST
-            be constructed with a leading zero-equals (<code>0=</code>) string, the string <code>ocfl_object_</code>,
-            followed by the OCFL specification version number. For example <code>0=ocfl_object_1.0</code> for version
-            1.0 of this specification.
+            The version declaration MUST be formatted according to the [[!NAMASTE]] specification. It MUST be an
+            file in the base directory of the <a>OCFL Storage Root</a> giving the OCFL version in the filename.
+            The filename MUST conform to the pattern <code>T=dvalue</code>, where <code>T</code> MUST be 0, and
+            <code>dvalue</code> MUST be <code>ocfl_object_</code>, followed by the OCFL specification version number.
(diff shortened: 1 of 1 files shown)


ahankinson:
I wasn't sure how to otherwise talk about the "value of the filename that comes after the 'Number Equals' bit" in a succinct way.

draft/spec/index.html
```diff
@@ -349,11 +349,12 @@ <h2>Object Structure</h2>
     <section id="object-conformance-declaration">
         <h2>Object Conformance Declaration</h2>
         <p>
-            The version declaration MUST be formatted according to the [[!NAMASTE]] specification. It MUST be an empty
-            file in the base directory of the object giving the OCFL Object version in the filename. The filename MUST
-            be constructed with a leading zero-equals (<code>0=</code>) string, the string <code>ocfl_object_</code>,
-            followed by the OCFL specification version number. For example <code>0=ocfl_object_1.0</code> for version
-            1.0 of this specification.
+            The version declaration MUST be formatted according to the [[!NAMASTE]] specification. It MUST be an
+            file in the base directory of the <a>OCFL Storage Root</a> giving the OCFL version in the filename.
```
awoods: Given the current section title, "Object Conformance Declaration", I would suggest not changing "base directory of the object" to "base directory of the OCFL Storage Root".
...unless we are no longer specifying the inclusion of a NAMASTE file in the object base.

rosy1280: 👍

rosy1280:
once changes @awoods recommends making are changed i approve.

draft/spec/index.html
zimeon: I think this is just a simple error, should be "OCFL Object" not "OCFL Storage Root"

ahankinson:
Refs #219 as well

ahankinson:
Errors have been fixed

ahankinson:
This can be merged now under the 72h rule.

Issue #249, opened 2018-10-19
Logical organisation of OCFL storage root


tomwrobel:
I appreciate that you've almost certainly gone over this many times, but I think it's still important to have some formal way of describing the logical organisation of the OCFL storage root.

I realise in spec/4.4 Extensions it says that you can provide the structure of the storage hierarchy, and in implementation/rebuildability there is a suggestion that this should be done. However, I don't think this goes far enough.

It might be made optional, but I think there should be an OCFL approved way to describe the storage hierarchy in relation to the object ID. 

Reasons:

1. It's good practice as this promotes long term sustainability. The storage hierarchy adopted may be obvious now, but that doesn't mean it will be so to future users of the data. An unexplained hierarchy is an additional problem future users will have to puzzle through.
2. Not doing so is a missed opportunity, as there is a chance to guide users towards a standard way to do this at the very start of the project.
3. It has practical benefits: In most implementations, I expect there would be a logical way to derive an object's location within the OCFL file system from its ID (otherwise an application would have to maintain a mapping externally). 
  3.1. Having a way to provide a formal definition of that relationship that is properly externally documented is very useful. 
  3.2 Although not necessarily in the first pass, an OCFL-approved statement (via NAMASTE?) could be used by a validation system to ensure that the Object ID/Object path relationship was intact.

Again, sorry if I'm being a pain, but this seemed to me to be something I actively want to be able to define in a compatible and sustainable way.

ahankinson:
Thanks Tom! You’re right, we have discussed this extensively; see #214 and #138. The gist of it is that it would certainly be a valuable thing to have, but that it would be better expressed in a complimentary spec, either on its own or as part of the pairtree spec.

The issues are thus: fundamentally that there is no satisfactory way to express content disposition succinctly with a namaste file. Embedding a regex in the filename to determine a pattern with which to validate and parse the paths will almost certainly run into problems with file names. Creating a naming scheme (e.g. 1=2-2-3-n) would be more preferable, but developing a naming scheme that covers all cases would be difficult. Some orgs will undoubtedly need to strip special characters (slashes, colons) from their paths, while others will be able to use them literally. Trying to encode all of these possibilities is beyond the scope of the OCFL effort, but the intent of the extensions mechanism is that if such a spec does emerge, OCFL will not get in the way.

Pairtree will also not be used in every implementation of OCFL. Object stores do not have the same limitations as traditional file systems, and storing them in a flat list is perfectly acceptable.

[...]

PR #268, opened 2018-11-07
Fixed: Clarify contents of OCFL Storage Root

### LINKED ISSUE #267: Clarify contents of Storage Root
From slack feedback:

"Is there a contradiction between "An OCFL Storage Root MUST NOT contain any files or directories other than those described here" and "The behavior of the storage root may be extended to support features from other specifications. An OCFL validator MUST ignore any files in the storage root it does not understand"? So, for instance, my pairtree_version0_1 file in the example above seems to run afoul of the first but is okay with the second."

ahankinson:
Fixes #267

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -745,11 +745,15 @@
           in the root.
         </p>
         <p>
-          An OCFL Storage Root MUST NOT contain any files or directories other than those described here.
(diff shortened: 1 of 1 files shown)


ahankinson:
Relates to #249

PR #269, opened 2018-11-07
Add versioning to NEED

### LINKED ISSUE #264: Add versioning to NEED
From e-mail feedback:

"Need: I was surprised that versioning wasn't mentioned in the Need statement, given that it is one of the key aspects of the spec and of its predecessor, Moab. It also is perhaps one of the most distinguishing differences from BagIt. I think this ends up underselling the spec a bit since having a standard way of structuring version directories is a key reason to adopt it. "

ahankinson:
Fixes #264 

Also a few minor fixes to the need section.

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -180,23 +180,23 @@
     <section id="need">
         <h2>Need</h2>
         <p>
-           The OCFL initiative arose from a need to have well-defined application-independent file management within
-           digital repositories.
+          The OCFL initiative arose from a need to have well-defined application-independent file management within
+          digital repositories.
         </p>
         <p>
-           A general observation is that the contents of a digital repository &#8212; that is, the digital files and
-           metadata that an institution might wish to manage &#8212; are largely stable. Once content has been
-           accessioned, it is unlikely to change significantly over its lifetime. This is in contrast to the software
-           applications that manage these contents, which are ephemeral, requiring constant updating and replacement.
-           Thus, transitions between application-specific methods of file management to support software upgrades and
-           replacement cycles can be seen as unnecessary and risky change which affects the long-term stable objects
-           in support of the short-term, ephemeral software.
+          A general observation is that the contents of a digital repository &#8212; that is, the digital files and
+          metadata that an institution might wish to manage &#8212; are largely stable. Once content has been
+          accessioned, it is unlikely to change significantly over its lifetime. This is in contrast to the software
+          applications that manage these contents, which are ephemeral, requiring constant updating and replacement.
+          Thus, transitions between application-specific methods of file management to support software upgrades and
+          replacement cycles can be seen as unnecessary and risky change which affects the long-term stable objects
+          in support of the short-term, ephemeral software.
         </p>
         <p>
-           By providing a specification for the file and directory layout on disk or in an object store, the OCFL is an
-           attempt at reducing, or even eliminating, the need for these transitions. As an application-independent
-           specification, conforming applications will natively "understand" the underlying file structure without
-           needing to first transition these contents to their own format.
+          By providing a specification for the file and directory layout on disk or in an object store, the OCFL is an
+          attempt at reducing, or even eliminating, the need for these transitions. As an application-independent
+          specification, conforming applications will natively "understand" the underlying file structure without
+          needing to first transition these contents to their own format.
          </p>
         <p>
           While digital objects are largely stable, there is also a need to track versions of these objects. These
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -199,10 +199,10 @@
           needing to first transition these contents to their own format.
          </p>
         <p>
-          While digital objects are largely stable, there is also a need to track versions of these objects. These
-          versions may appear due to bitstream format shifts, preservation interventions to address corruption, or
-          to fix errors discovered after accessioning. The OCFL attempts to rationalize the file structures necessary
-          to capture versions of object contents so that previous states of an object may be recovered and examined.
+          While digital repository content changes relatively slowly, it is necessary to be able to track changes to 
+          digital objects. Within the file and directory layout specification, OCFL provides a simple structure to 
+          efficiently capture versions of object contents so that all previous states of an object may be recovered
+          and examined.
         </p>
     </section>
 </section>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -180,24 +180,30 @@ <h2>Introduction</h2>
...
+          Thus, transitions between application-specific methods of file management to support software upgrades and
+          replacement cycles can be seen as unnecessary and risky change which affects the long-term stable objects
+          in support of the short-term, ephemeral software.
         </p>
         <p>
-           By providing a specification for the file and directory layout on disk or in an object store, OCFL is an
-           attempt at reducing, or even eliminating, the need for these transitions. As an application-independent
-           specification, conforming applications will natively "understand" the underlying file structure without
-           needing to first transition these contents to their own format.
+          By providing a specification for the file and directory layout on disk or in an object store, the OCFL is an
```
zimeon: I'm not sure that we want to say "the OCFL" everywhere. I'm fine with "the OCFL initiative" but would otherwise just say "OCFL" -- we should decide about this as a style issue as a change would also have impact on the implementation notes --> #270

draft/spec/index.html
```diff
@@ -180,24 +180,30 @@ <h2>Introduction</h2>
...
+          attempt at reducing, or even eliminating, the need for these transitions. As an application-independent
+          specification, conforming applications will natively "understand" the underlying file structure without
+          needing to first transition these contents to their own format.
          </p>
+        <p>
+          While digital objects are largely stable, there is also a need to track versions of these objects. These
```
zimeon: "digital objects are largely stable" does not seem to be a true statement to me in the general case. It seems that it would be true to for the archival use case so we could say:

> While archival digital object are expected to be largely stable, ...

but we haven't restricted OCFL to this use case elsewhere. In fact, later we say:

> A central feature of the OCFL specification is support for versioning. This recognizes that digital objects will change over time, through new requirements, fixes, updates, or format shifts. The specification takes no position on what constitutes a version or a versionable action, but it is recommended that implementers have a clear position on this within their local storage policies.

So perhaps we should just say:

> While in many applications digital objects are largely stable, there is also a need to track versions of these objects. 

and then remove the following sentence which is too limited to the archival case and not aligned with the later statement about possible reasons for versions.

ahankinson: We do start the "Need" section with:

> A general observation is that the contents of a digital repository -- that is, the digital files and metadata that an institution might wish to manage -- are largely stable.

ahankinson: So I was trying to work within that context for the digital object, since it follows that if a repository is largely stable, the objects in that repository must also be largely stable.

zimeon: I still feel that the earlier comment is a bit of an over statement, but it is a nice motivation for file and directory layout. Perhaps the key distinction here is that the **files** are largely stable (hence it is sensible to do deltas by file) but **objects** are more frequently added and added to, although we also support arbitrary updates. I'll make a new suggestion in the main thread as it touches on all three of the bits I commented on...

zimeon:
Here is my attempt at an alternative suggestion based on the discussion [above](https://github.com/OCFL/spec/pull/269#pullrequestreview-172516683) with @ahankinson. I don't think we should get into the reasons for changes as we do that [later on (end of the Object intro)](https://ocfl.io/draft/spec/#object-spec):

> While digital repository content changes relatively slowly, it is necessary to be able to track changes to digital objects. Within the file and directory layout specification, OCFL provides a simple structure to efficiently capture versions of object contents so that all previous states of an object may be recovered and examined.

ahankinson:
Why? They're easily reviewed since they are in a separate commit.

awoods:
There are four commits in this PR. Normally I would squash them in order to have one clear commit that addresses the referenced issue. Are you suggesting keeping these four interim commits in the final history?

ahankinson:
We can squash them; that’s fine. I just don’t see the need for separating some minor formatting commits for the same section into its own PR. 

If it were a massive set of unrelated formatting commits, then that’s an issue. But I really don’t see the need to break these out into a separate PR.

awoods:
I find it easier to do reviews and future github forensics with clean commits that are isolated to the associated issues. I find unrelated formatting updates distracting and time-consuming (I have to look at each line to see if there is actually a functional change). 

I would be happy to create a PR against this branch to remove the formatting changes... or if you like them for your own reasons, we can go forward as-is.

draft/spec/index.html
```diff
@@ -180,24 +180,30 @@ <h2>Introduction</h2>
...
+          By providing a specification for the file and directory layout on disk or in an object store, the OCFL is an
+          attempt at reducing, or even eliminating, the need for these transitions. As an application-independent
+          specification, conforming applications will natively "understand" the underlying file structure without
+          needing to first transition these contents to their own format.
          </p>
+        <p>
+          While digital repository content changes relatively slowly, it is necessary to be able to track changes to 
+          digital objects. Within the file and directory layout specification, OCFL provides a simple structure to 
+          efficiently capture versions of object contents so that all previous states of an object may be recovered
+          and examined.
```
rosy1280: Should we also note that the specification makes no recommendations on what constitutes a version, that is left to the institution to decide.

zimeon: I don't think we need to put that in the need section -- we already have it later https://ocfl.io/draft/spec/#object-spec

> A central feature of the OCFL specification is support for versioning. This recognizes that digital objects will change over time, through new requirements, fixes, updates, or format shifts. The specification takes no position on what constitutes a version or a versionable action, but it is recommended that implementers have a clear position on this within their local storage policies.

rosy1280: I'll approve and contemplate more language here then.  While we talk a great deal in this section about the need for application independent storage, this paragraph in contrast seems weak given so many folks see versioning as the strength over something like BagIT (which could also be used as application independent storage and often is).

ahankinson:
@awoods caveat the formatting changes, are there any further changes to the content of the PR that you would like to see?

awoods:
@ahankinson : No other suggestions. The content looks good.

awoods:
@ahankinson : here is a whitespace-cleanup PR against your branch: https://github.com/OCFL/spec/pull/276

PR #286, opened 2018-12-05
Fixed: Clarify that BagIt example is only one way

### LINKED ISSUE #266: Bagit Example
From e-mail feedback:

"5.3 BagIt as an OCFL Object: When resolving bags from OCFL objects, it appears that the inventory.json files are not included in the resulting bag. The inventory has potentially useful version information in it and it seems like you'd want that as a tag file in some form."

Need to figure out whether we will address this.

ahankinson:
Fixes #266


ahankinson:
This can be merged now under the 72h rule.

PR #294, opened 2018-12-31
Fixed: Clarify inventory.json must be present

### LINKED ISSUE #293: Clarify that there MUST be an inventory.json file in the Object Root
We have decided to make the presence of an inventory.json file optional in the version directories (3.7 "Every version directory SHOULD include an inventory file that is an Inventory of all content for versions up to and including that particular version.") However, we don't also specify a MUST for the inventory.json file in the root of the object. 

This can lead to the absurd situation where an OCFL Object without an inventory.json file is technically valid! 

Propose modifying §3.7 to read:

"Every OCFL Object MUST have an inventory file within the OCFL Object Root, corresponding to the state of the OCFL Object at the current version. Additionally, every version directory SHOULD include an inventory file that is an Inventory of all content for versions up to and including that particular version. Where an OCFL Object contains `inventory.json` in version directories, the `inventory.json` in the OCFL Object Root MUST be the same as the file in the most recent version."

ahankinson:
Fixes #293

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -704,9 +704,13 @@
     <section id="version-inventory">
         <h2>Version Inventory and Inventory Digest</h2>
         <p>
-            Every version directory SHOULD include an inventory file that is an <a href="#inventory">Inventory</a>
-            of all content for versions up to and including that particular version. Every inventory file MUST have
-            a corresponding <a href="#inventory-digest">Inventory Digest</a>.
+            Every OCFL Object MUST have an  inventory file within the OCFL Object Root,
+            corresponding to the state of the OCFL Object at the current version. Additionally, every version directory
+            SHOULD include an inventory file that is an <a href="#inventory">Inventory</a> of all content for versions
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -704,9 +704,13 @@ <h2>Inventory Digest</h2>
     <section id="version-inventory">
         <h2>Version Inventory and Inventory Digest</h2>
         <p>
-            Every version directory SHOULD include an inventory file that is an <a href="#inventory">Inventory</a>
-            of all content for versions up to and including that particular version. Every inventory file MUST have
-            a corresponding <a href="#inventory-digest">Inventory Digest</a>.
+            Every OCFL Object MUST have an  inventory file within the OCFL Object Root,
```
awoods: Minor: extraneous space in `have an  inventory file`.
Otherwise: :+1:

zimeon: What @awoods said...

ahankinson:
Fixéd.

PR #298, opened 2019-02-06
Remove ocfl-tech@googlegroups.com from README

awoods:


zimeon:
This is a minor change so next :+1: can merge with not waiting period

PR #309, opened 2019-02-18
Add note about zero padding not being recommended

### LINKED ISSUE #304: Add Unix `ls` discussion to implementation-notes
Split from #281: we have agreed to add a discussion of zero padding and lexical sorting in Unix `ls` to the [implementation-notes](https://ocfl.io/draft/implementation-notes/). See also #2. I propose something like the following as a new paragraph to [3.2.1 Version Numbering](https://ocfl.io/draft/implementation-notes/#version-numbering):

> Systems with version directories have often used zero-padding in order to show version order with common lexical sorting tools (such as Unix `ls`). Zero-padding is not recommended in order to avoid having to make arbitrary choices of padding length or to place limits of the number of versions supported. In archival storage use cases, objects with more than 9 versions are rare, and while it is important that manual OCFL object be possible, it is not expected to be the norm. OCFL-aware tools can, of course, easily sort unpadded version directories in version order.

zimeon:
Fixes #304

### DIFF
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -492,7 +492,7 @@
         <p>
           Systems with version directories have often used zero padding in order to show version order with common
           lexical sorting tools (such as Unix <code>ls</code>). Zero padding is not recommended in order to avoid
-          having to make arbitrary choices of padding length or to place limits of the number of versions supported.
+          having to make arbitrary choices of padding length or to place limits on the number of versions supported.
         </p>
       </section>
 
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -489,6 +489,11 @@
           following <code>v</code> is a zero then the number format is zero padded to fixed length, otherwise it
           is simply an integer.
         </p>
+        <p>
+          Systems with version directories have often used zero padding in order to show version order with common
+          lexical sorting tools (such as Unix <code>ls</code>). Zero padding is not recommended in order to avoid
+          having to make arbitrary choices of padding length or to place limits of the number of versions supported.
+        </p>
       </section>
 
       <section>


draft/implementation-notes/index.html
```diff
@@ -489,6 +489,11 @@ <h2>Version Numbering</h2>
           following <code>v</code> is a zero then the number format is zero padded to fixed length, otherwise it
           is simply an integer.
         </p>
+        <p>
+          Systems with version directories have often used zero padding in order to show version order with common
+          lexical sorting tools (such as Unix <code>ls</code>). Zero padding is not recommended in order to avoid
+          having to make arbitrary choices of padding length or to place limits of the number of versions supported.
```
awoods: typo?
Should the `of` be `on`:
"place limits [of] the number of versions"

zimeon: Indeed, will fix

draft/implementation-notes/index.html
```diff
@@ -489,6 +489,11 @@ <h2>Version Numbering</h2>
           following <code>v</code> is a zero then the number format is zero padded to fixed length, otherwise it
           is simply an integer.
         </p>
+        <p>
```
ahankinson: Perhaps:

"In the OCFL, zero padding is not recommended to avoid having to make arbitrary choices of padding length or to place limits on the number of versions sorted. However, on most operating systems this has the effect of showing directories sorted in lexical order, (where "10" comes before "2") rather than natural, numerical order. Numeric sorting may instead be accomplished with the following commands:
 - `ls | sort -n` (most systems)
 - `ls -v` (GNU Linux)
 - `dir | sort` (Windows Command prompt)"
---

zimeon: That seems like too much detail to me, we don't go to that level anywhere else. Also, I don't think those commands work for filenames with with the preceding `v`

ahankinson: The implementation notes are (I thought) where details like this would live. "It doesn't do lexical sorting by default *as written in the spec* but here's how you can achieve this."

The commands work for me (slight adjustment to `sort` args):

```
mac$ ls
v1   v10  v100 v2   v20  v3
mac$ ls | sort -n -k1.2
v1
v2
v3
v10
v20
v100
---
linux$ $ ls
v1  v10  v100  v2  v20  v3
linux$ ls -v
v1  v2  v3  v10  v20  v100
---
(Volunteers to test in Windows?) 
```

zimeon: Cool that adding `-k1.2` fixes it for sort, I didn't know about that option.

I think we need to have a editors' discussion about the level of detail though. I'm not opposed the idea of providing such detail but we don't yet do it anywhere and it would have implications for the maintenance of the implementation notes document.

awoods: My first reaction was similar to yours, @zimeon : the suggestion seemed heavy on low-level details.

If we consider the implementation-notes as a scratch-pad for related ideas, the detailed approach is probably ok. My inclination is to keep the implementation-notes a bit more general... but I don't think the alternative is a critical issue.

rosy1280: I think we should keep it simple, right now the detail feels like here's how you do unix ls

ahankinson: OK, I can withdraw my review, reluctantly, but with reservations.

I think we all agree that for OCFL implementations, the devil will be in the implementation details. It's all well and good to define a spec for the content layout, but the *risk* will always be with content processing. 

The idea for the implementation notes, as I remember it, was as a place to put our thinking about interactions with "the object in motion."

Without specific, concrete interaction examples, people are going to produce valid OCFL Objects, but how they get there will be uncontrolled. I'm not saying we need to be prescriptive about how people have to implement an OCFL client library, but I am saying that we, at least, should give people examples of how to do something so that they can either copy-and-paste, or have an understanding of how they might be deviating from the intended design.

If we're couching the implementation notes in the same abstract language as the spec, is there any other place where we can give people real, concrete examples of how to interact with an OCFL Object?

awoods:
Thanks for your thoughtfulness on this, @ahankinson . I completely agree with your sentiment of using the implementation-notes as a home for clarifying guidance. 

For me, the suggestion in this PR went too far by detailing how to use `ls`. Maybe a compromise would be a statement indicating that "common sorting tools can be configured to perform numeric sorting without zero-padding".

The current wording of this PR is along the lines of:
- Some people use zero padding for easy lexical sorting
- Zero padding is not recommended due to the limitations it imposes

..maybe a third phrase could be added (in a follow-on PR):
- "common sorting tools can be configured to perform numeric sorting without zero-padding"

Issue #285, opened 2018-12-05
Case sensitivity and OCFL


rosy1280:
Should go in the Spec. See: https://github.com/OCFL/spec/pull/227#discussion-diff-224908719R375

awoods:
What is the exact suggestion in this issue?

neilsjefferies:
OCFL requires case sensetivity and case preservation. Not supported on file systems that can't manage that.

Hard statement in spec and admit that it can be done with caveats in Imp Notes.

zimeon:
This does not appear to have been addressed in #317. We could add a bullet point to [Filesystem Feature](https://ocfl.io/draft/spec/#filesystem-features) along the lines of:

|  * File paths and filenames in the OCFL are cases sensitive. Filesystems MUST preserve the case of OCFL file paths and filenames.

Issue #284, opened 2018-12-05
UTF Encoding for OCFL


rosy1280:
see: https://github.com/OCFL/spec/pull/227#discussion-diff-224943314R374 for more discussion

zimeon:
Proposal from editors' call. In the specification we will add comment that:
  1. The inventory, as JSON, is in UTF
   2. The OCFL relies upon a storage layer (filesystem or object store) that consistently and unambiguously maps unicode pathnames in the inventory files/objects.

zimeon:
Editors' call: #317 didn't add anything explicit about UTF but we have decided that we don't need to mention it explicitly. JSON is already clearly defined over UTF

PR #317, opened 2019-03-06
Filesystem features in implementation notes

neilsjefferies:
#284 and #285

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -884,7 +884,7 @@
               MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
               Unicode which MUST be used for all OCFL-related operations.</li>
           <li>The OCFL REQUIRES that filesystems are both case-preserving and case-sensitive so that the mapping 
-              between storage paths in the inventory and the filesystem is consistent and unqiue.</li>
+              between storage paths in the inventory and the filesystem is consistent and unique.</li>
           <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
               filesystems or preservable through file transfer operations. Nor can these attributes be validated in 
               terms of fixity in a consistent manner. As such, the OCFL does NOT support the portability of these
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -872,28 +872,28 @@
     <section id="filesystem-features">
         <h2>Filesystem features</h2>
         <p>
-            In order to maximise the compatibility of the OCFL with different filesystems, and thus improve the 
-            portability of OCFL Objects betwwen different systems, some restrictions on the use of certain 
+            In order to maximize the compatibility of the OCFL with different filesystems, and thus improve the
+            portability of OCFL Objects between different systems, some restrictions on the use of certain
             filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
-            data MUST be wrappered in a suitable disk or filesystem image format which OCFL can treat as a regular
+            data MUST be wrapped in a suitable disk or filesystem image format which OCFL can treat as a regular
             file.
         </p>
         <ol>
-          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the 
+          <li>The use of JSON for the inventory means that content paths are stored as Unicode. Therefore, the
               filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
               MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
               Unicode which MUST be used for all OCFL-related operations.</li>
-          <li>The OCFL REQUIRES that filesystems are both case-preserving and case-sensitive so that the mapping 
-              between storage paths in the inventory and the filesystem is consistent and unique.</li>
+          <li>OCFL storage MUST be both case-preserving and case-sensitive so that the mapping
+              between content paths in the inventory and the filesystem is consistent and unique.</li>
           <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
-              filesystems or preservable through file transfer operations. Nor can these attributes be validated in 
-              terms of fixity in a consistent manner. As such, the OCFL does NOT support the portability of these
+              filesystems or preservable through file transfer operations. Nor can these attributes be validated in
+              terms of fixity in a consistent manner. As such, the OCFL does not support the portability of these
               attributes.</li>
           <li>Hard and soft (symbolic) links are not portable and MUST NOT be used within OCFL Storage hierachies.
               A common use case for links is storage deduplication. OCFL inventories provide a portable method
               of achieving the same effect.</li>
-          <li>Transparent filesystem features such as compression and encryption should be effectively invisible to 
-              OCFL operations. Consequently, they should NOT be expected to be portable.</li>
+          <li>Transparent filesystem features such as compression and encryption should be effectively invisible to
+              OCFL operations. Consequently, they should not be expected to be portable.</li>
         </ol>
     </section>
 
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -879,12 +879,6 @@
             file.
         </p>
         <ol>
-          <li>The use of JSON for the inventory means that content paths are stored as Unicode. Therefore, the
-              filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
-              MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
-              Unicode which MUST be used for all OCFL-related operations.</li>
-          <li>OCFL storage MUST be both case-preserving and case-sensitive so that the mapping
-              between content paths in the inventory and the filesystem is consistent and unique.</li>
           <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
               filesystems or preservable through file transfer operations. Nor can these attributes be validated in
               terms of fixity in a consistent manner. As such, the OCFL does not support the portability of these
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -868,6 +868,34 @@
             </p>
         </blockquote>
     </section>
+    
+    <section id="filesystem-features">
+        <h2>Filesystem features</h2>
+        <p>
+            In order to maximise the compatibility of the OCFL with different filesystems, and thus improve the 
+            portability of OCFL Objects betwwen different systems, some restrictions on the use of certain 
+            filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
+            data MUST be wrappered in a suitable disk or filesystem image format which OCFL can treat as a regular
+            file.
+        </p>
+        <ol>
+          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the 
+              filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
+              MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
+              Unicode which MUST be used for all OCFL-related operations.</li>
+          <li>The OCFL REQUIRES that filesystems are both case-preserving and case-sensitive so that the mapping 
+              between storage paths in the inventory and the filesystem is consistent and unqiue.</li>
+          <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
+              filesystems or preservable through file transfer operations. Nor can these attributes be validated in 
+              terms of fixity in a consistent manner. As such, the OCFL does NOT support the portability of these
+              attributes.</li>
+          <li>Hard and soft (symbolic) links are not portable and MUST NOT be used within OCFL Storage hierachies.
+              A common use case for links is storage deduplication. OCFL inventories provide a portable method
+              of achieving the same effect.</li>
+          <li>Transparent filesystem features such as compression and encryption should be effectively invisible to 
+              OCFL operations. Consequently, they should NOT be expected to be portable.</li>
+        </ol>
+    </section>
 
 </section>
 
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -875,17 +875,17 @@
             In order to maximize the compatibility of the OCFL with different filesystems, and thus improve the
             portability of OCFL Objects between different systems, some restrictions on the use of certain
             filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
-            data MUST be wrapped in a suitable disk or filesystem image format which OCFL can treat as a regular
+            the content MUST be wrapped in a suitable disk or filesystem image format which OCFL can treat as a regular
             file.
         </p>
         <ol>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
         </blockquote>
     </section>
+    
+    <section id="filesystem-features">
+        <h2>Filesystem features</h2>
+        <p>
+            In order to maximise the compatibility of the OCFL with different filesystems, and thus improve the 
+            portability of OCFL Objects betwwen different systems, some restrictions on the use of certain 
+            filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
+            data MUST be wrappered in a suitable disk or filesystem image format which OCFL can treat as a regular
```
ahankinson: Add (e.g., ISO or Tar)?

zimeon: wrappered or simply wrapped? Or would encapsulated be better?

If we say "tar" I think all lowercase is conventional (see, e.g. https://en.wikipedia.org/wiki/Tar_(computing) ). I'm +0 on adding these examples, it feels a little awkward to add things like that without good links to specs etc. and without consideration of what bias we might be inserting into choice of format.

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
+            file.
+        </p>
+        <ol>
+          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the
```
zimeon: s/in as/as/

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
+            filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
+            data MUST be wrappered in a suitable disk or filesystem image format which OCFL can treat as a regular
+            file.
+        </p>
+        <ol>
+          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the 
+              filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
+              MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
+              Unicode which MUST be used for all OCFL-related operations.</li>
+          <li>The OCFL REQUIRES that filesystems are both case-preserving and case-sensitive so that the mapping
```
zimeon: _REQUIRES_ is not one of the [RFC2119 words we have chosen to use. I suggest:

> OCFL storage MUST be both ...

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
+        <ol>
+          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the 
+              filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
+              MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
+              Unicode which MUST be used for all OCFL-related operations.</li>
+          <li>The OCFL REQUIRES that filesystems are both case-preserving and case-sensitive so that the mapping 
+              between storage paths in the inventory and the filesystem is consistent and unique.</li>
+          <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
+              filesystems or preservable through file transfer operations. Nor can these attributes be validated in 
+              terms of fixity in a consistent manner. As such, the OCFL does NOT support the portability of these
```
zimeon: s/NOT/not/

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
+              attributes.</li>
+          <li>Hard and soft (symbolic) links are not portable and MUST NOT be used within OCFL Storage hierachies.
+              A common use case for links is storage deduplication. OCFL inventories provide a portable method
+              of achieving the same effect.</li>
+          <li>Transparent filesystem features such as compression and encryption should be effectively invisible to 
+              OCFL operations. Consequently, they should NOT be expected to be portable.</li>
```
zimeon: s/NOT/not/  since it is talking about expectations rather than what one "SHOULD NOT" (all caps) do

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
...
+            In order to maximise the compatibility of the OCFL with different filesystems, and thus improve the 
+            portability of OCFL Objects betwwen different systems, some restrictions on the use of certain 
+            filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
+            data MUST be wrappered in a suitable disk or filesystem image format which OCFL can treat as a regular
+            file.
+        </p>
+        <ol>
+          <li>The use of JSON for the inventory means that storage paths are stored in as Unicode. Therefore, the 
+              filesystem hosting an OCFL Storage Root SHOULD support Unicode file- and path-names. If not, there
+              MUST be a consistent, symmetric mapping between the character set used for filesystem paths and
```
birkland: I'm not quite sure I understand what this requirement is saying.  If the storage filesystem does not support unicode, then isn't it simply the case that the "content paths" (which I believe is the new terminology for what was formerly known as "existing file path") are subject to the limitations of the local filesystem, and therefore cannot contain unicode on filesystems that don't support it?  

Additionally, it's not entirely clear where/how this mapping occurs.  When transferring an OCFL object between systems with different filesystem naming capabilities, is this mapping referring to software that might need to re-write the filenames and update the `inventory.json` files to account for local filesystem limitations?  i.e. _there MUST be a consistent, symmetric mapping..._ is placing constraints on the process of migrating an OCFL objects between filesystems of different capabilities?

draft/spec/index.html
julianmorley: s/wrappered/wrapped

draft/spec/index.html
```diff
@@ -868,6 +868,34 @@ <h2>Extensions</h2>
             </p>
         </blockquote>
     </section>
+    
+    <section id="filesystem-features">
+        <h2>Filesystem features</h2>
+        <p>
+            In order to maximise the compatibility of the OCFL with different filesystems, and thus improve the 
+            portability of OCFL Objects betwwen different systems, some restrictions on the use of certain
```
julianmorley: s/betwwen/between

draft/spec/index.html
ahankinson: I think the issue we've identified is where objects traverse filesystems that may have different character encodings for their filesystems. A filesystem can support Unicode, but not in the same way that OTHER filesystems support it.

HFS+, for example, treats Unicode in a different way from most other modern filesystems, using a modified variant of NFD. 

https://en.wikipedia.org/wiki/HFS_Plus#Design

APFS, on the other hand, does not use any Unicode normalization:

https://mjtsai.com/blog/2017/03/24/apfss-bag-of-bytes-filenames/

This can cause problems:

https://eclecticlight.co/2017/04/06/apfs-is-currently-unusable-with-most-non-english-languages/

APFS isn't the only one:

https://gist.github.com/Kwpolska/0c63a22f65efe9c90607a826aad99d1f

I think fundamentally we just don't have enough data to make a real recommendation here, so we're hedging our bets.

birkland: I believe I understand the motivating issue.  The problem is that the language is not clear as to what "consistent, symmetric mapping" means.  Where does the mapping occur, who is performing the mapping, and what implication does it have for the contents of the `inventory.json` file if any?

To use an HFS+ example, the filesystem itself performs a "mapping" (of é to the three bytes 65 cc 81, which may not be equivalent to the bytes used for é in the inventory storage paths.).  Is this clause just saying that it is OK for the to use OCFL with filesystems that do that?  Or is it allowing (or requiring) OCFL software to implement any sort of workaround it wants in order to deal with `inventory.json` and filesystem path diffferences, as long as it involves a "a consistent, symmetric mapping"?  

Absent the mapping clause, my interpretation is that the storage paths used in the `inventory.json` file would be subject to the local filesystem costraints.  i.e. if your FS requires é to be `65 cc 81`, those bytes must be in the storage path in the inventory, no exceptions.  I thought this is where the flexibity afforded by logical paths comes in.  If the desired name of the file has é as `c3 a9`, it's fine to have that normalization in the logical file path, but if you happen to be using HFS+ to persist OCFL, your hands are tied to using  `65 cc 81` in the storage path.

ahankinson: I think we're trying to deal with a case where an OCFL object written on Filesystem A using `65 cc 81` writes an inventory, and then the whole lot gets moved to Filesystem B, where the file is understood as `c3 a9`. In this case, as I understand it, the inventory would contain the bytes `65 cc 81` (because it doesn't get rewritten) but a client attempting to look that path up on the filesystem wouldn't find it. This might happen if, for example, you copy a file from a local filesystem to an Object Store.

Like I said, we don't really have a whole lot of use cases to know if this is an actual worry, or just one that we're imagining.

birkland: I believe filesystems that do not preserve case (i.e. the FS operations normalize lowercase characters to uppercase) are an extreme example of the same phenomenon, and may be more tangible https://github.com/OCFL/spec/issues/285

On Windows, it's even worse.  Characters like `:`,`<`,`?`,`*` are prohibited entirely, as are character sequences like `NUL`.   The filesystem throws an error, rather than re-mapping a character automatically.  The mapping clause might be interpreted as allowing an implementation to re-name the files using its own convention, as long as it's consistent with the rules it applies to do so.  Is this accurate?  

I worry that there is a point at which flexibility in the name of portability can complicate the archival characteristics of OCFL.

ahankinson: It's one of the reasons why we're recommending that if you do want to do something with older filesystems, ISO or tar formats may be useful. Especially when it comes to archiving full disk images with empty directories.

birkland: Fair enough.  As an outsider, it's easier to understand the idea that portability might not be possible across filesystems, and there is nothing that can be done about that aside from judicious choices of storage paths (file names) in the first place, or wrapping it in an archival storage format like a tar.  So the first paragraph of this PR is fine, it's just that the normative remediation for filesystem path inconsistencies (due to character encoding, support for ASCII case, unsupported characters, etc) is something else completely: a mapping of some sort, the nature of which is undefined.  Why wouldn't path issues fall under the purvue of "if portability problems arise, you need to wrap" as well?

neilsjefferies: ...This should probably be dealt with in the Implementation Notes rather than here.

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
             </p>
         </blockquote>
     </section>
+    
+    <section id="filesystem-features">
+        <h2>Filesystem features</h2>
+        <p>
+            In order to maximize the compatibility of the OCFL with different filesystems, and thus improve the
```
ahankinson: remove `thus`

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
...
+            portability of OCFL Objects between different systems, some restrictions on the use of certain
+            filesystem features are necessary. If the preservation of non-OCFL-compliant features is required then
+            data MUST be wrapped in a suitable disk or filesystem image format which OCFL can treat as a regular
```
ahankinson: 'data' => 'the contents' (to make it more consistent with how we refer to the content of an OCFL object)

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
...
+            file.
+        </p>
+        <ol>
+          <li>Filesystem metadata (e.g. permissions, access, and creation times) are not considered portable between
+              filesystems or preservable through file transfer operations. Nor can these attributes be validated in
```
ahankinson: 'Nor can' => 'These attributes also cannot be validated ... '

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
...
+              terms of fixity in a consistent manner. As such, the OCFL does not support the portability of these
+              attributes.</li>
+          <li>Hard and soft (symbolic) links are not portable and MUST NOT be used within OCFL Storage hierachies.
```
ahankinson: Non-acceptable RFC language...

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
...
+              A common use case for links is storage deduplication. OCFL inventories provide a portable method
```
ahankinson: This seems to start making a point and then drifts off.

draft/spec/index.html
```diff
@@ -868,6 +868,28 @@ <h2>Extensions</h2>
...
+              of achieving the same effect.</li>
+          <li>Transparent filesystem features such as compression and encryption should be effectively invisible to
+              OCFL operations. Consequently, they should not be expected to be portable.</li>
```
ahankinson: To what effect?

zimeon: @ahankinson - are you suggesting we remove this bullet entirely?

draft/spec/index.html
zimeon: The point seems succinct to me. Are you suggesting we add something along the lines of "and so links are not necessary"?

draft/spec/index.html
zimeon: ?? MUST NOT is noted in https://ocfl.io/draft/spec/#conformance -- is there something else?

draft/spec/index.html
zimeon: :+1:

draft/spec/index.html
zimeon: or "the content" ? (To say "the contents" begs "of what?" to me)

draft/spec/index.html
zimeon: maximizing compatibility does improve portability so the "thus" seems appropriate to me

draft/spec/index.html
ahankinson: Sure; I just wanted to make sure we were being consistent, since we are trying to talk about both data and metadata as the 'content' of an OCFL Object.

draft/spec/index.html
ahankinson: My bad.

draft/spec/index.html
ahankinson: No; I'm saying that we should expand a bit on what that could mean. Moving from a compressed to an uncompressed filesystem may significantly increase the size of your data. Likewise, simply cloning an encrypted disk may make it unreadable. Simply saying they are not portable hides more important issues IMO.

draft/spec/index.html
ahankinson: IMO it would be useful to know how inventories achieve the same effect; that's not spelled out in the spec, so people are left to fill in the gaps.

rosy1280: Would language like:

> OCFL inventories provide a portable method of achieving the same effect by using hashes to address content.

zimeon: s/hashes/digests/ but yes

draft/spec/index.html
rosy1280: @ahankinson do you have language to suggest?  I'm struggling with finding succinct sentences.

Issue #313, opened 2019-02-23
Consider changing existing file path to something else


rosy1280:
I've noticed that folks continue to call existing file path by another name.  Should we consider changing the term to storage file path or physical file path.

ahankinson:
Storage file path is a possibility. “Physical” implies an actual physical object to me, which is perhaps not accurate in a digital context.

rosy1280:
Agreed. I note physical because I keep seeing folks refer to it that way. But then in the article @dbrower used storage path which seemed like a better fit.

zimeon:
I see _physical_ as the usual opposite of _logical_ but I agree with @ahankinson that it seems to conjure the wrong impression. On balance I like **storage path** and **logical path**, and prefer them without the _file_ because then we avoid always sounding like a filesystem as opposed to some other storage system (like S3), and avoid saying file twice in various sentences. To take an [example sentence](https://ocfl.io/draft/spec/#manifest), we might then change:

> The value for each key is an array containing the **existing file paths** of files in the OCFL Object that have content with the given digest.

to

> The value for each key is an array containing the **storage paths** of files in the OCFL Object that have content with the given digest.

birkland:
"storage path" seems like a more universally understood metaphor between the world of byte wrangling software engineers and librarians/archivists.   It reads well in the proposed language, and aligns well with the term "storage root"

awoods:
+1 to "storage path"

ahankinson:
Does it align too well with "storage root"? As in, are people likely to get confused that there is the storage root, and then a bunch of stuff in the middle that doesn't have a name, and then an object root, and then within the object root... a storage path?

rosy1280:
@ahankinson that's a great point, but i think on the whole, folks are more likely to not use the storage root terminology as frequently as the storage path terminology (based on documents i've seen and conversations i've had so far).

ahankinson:
Editors call 2019-03-06: Debated storage path terminology.

Decision: "existing file path" should change to "content path" in the documentation, and "logical file path" should be normalized to "logical path".

zimeon:
This still needs a PR to make the terminology changes in the implementation notes (#333 fixed in spec)

PR #321, opened 2019-03-15
Add renaming example

### LINKED ISSUE #310: Add inventory example to implementation notes that has different existing and logical paths
Based on https://github.com/OCFL/spec/issues/275#issuecomment-463774897 it would make things clearer to have an example that has different existing and logical paths

zimeon:
Fixes #310. Adds "5.3 File Renaming in an OCFL Object".

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -1031,6 +1031,43 @@
 </pre>
     </section>
 
+    <section id="example-object-with-file-renaming">
+        <h2>File Renaming in an OCFL Object</h2>
+        <p>
+            The following example OCFL Object inventory shows how exiting file paths may differ from
+            logical file paths. The example object has just one version, <code>v1</code>, which has
+            two files with logical file paths <code>.hidden.txt</code> and
+            <code>name with spaces.txt</code> as shown in the <code>state</code> block. These names
+            have been normalized to avoid spaces and leading period in the existing file paths. The
+            <code>manifest</code> block shows <code>%2Ehidden.txt</code> and
+            <code>name%20with%20spaces.txt</code> within the <code>v1/content</code> directory. This
+            specification does not constrain the normalization or renaming strategy adopted, in
+            this example URI encoding was used but many other approaches are possible, and they may
+            or may not yield an obvious relation between exiting file paths and logical file paths.
+        </p>
+<pre>
+{
+  "digestAlgorithm": "sha512",
+  "head": "v1",
+  "id": "http://example.org/rename",
+  "manifest": {
+    "2deeaf...c0f": [ "v1/content/%2Ehidden.txt" ],
+    "329ebf...a6a": [ "v1/content/name%20with%20spaces.txt" ]
+  },
+  "type": "https://ocfl.io/1.0/spec/#inventory",
+  "versions": {
+    "v1": {
+      "created": "2019-03-14T20:31:00Z",
+      "state": {
+        "2deeaf...c0f": [ ".hidden.txt" ],
+        "329ebf...a6a": [ "name with spaces.txt" ]
+      }
+    }
+  }
+}
+</pre>
+    </section>
+
     <section id="example-bagit-in-ocfl">
         <h2>BagIt in an OCFL Object</h2>
         <p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -1031,36 +1031,38 @@
 </pre>
     </section>
 
-    <section id="example-object-with-file-renaming">
-        <h2>File Renaming in an OCFL Object</h2>
+    <section id="example-object-diff-paths">
+        <h2>Different Logical and Existing File Paths in an OCFL Object</h2>
         <p>
             The following example OCFL Object inventory shows how existing file paths may differ from
             logical file paths. The example object has just one version, <code>v1</code>, which has
-            two files with logical file paths <code>.hidden.txt</code> and
-            <code>name with spaces.txt</code> as shown in the <code>state</code> block. These names
-            have been normalized to avoid spaces and leading period in the existing file paths. The
-            <code>manifest</code> block shows <code>%2Ehidden.txt</code> and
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -1031,6 +1031,43 @@ <h2>Versioned OCFL Object</h2>
 </pre>
     </section>
 
+    <section id="example-object-with-file-renaming">
+        <h2>File Renaming in an OCFL Object</h2>
+        <p>
+            The following example OCFL Object inventory shows how exiting file paths may differ from
```
ahankinson: `exiting [sic] file paths` => `storage paths`

draft/spec/index.html
```diff
@@ -1031,6 +1031,43 @@ <h2>Versioned OCFL Object</h2>
...
+            logical file paths. The example object has just one version, <code>v1</code>, which has
+            two files with logical file paths <code>.hidden.txt</code> and
+            <code>name with spaces.txt</code> as shown in the <code>state</code> block. These names
+            have been normalized to avoid spaces and leading period in the existing file paths. The
+            <code>manifest</code> block shows <code>%2Ehidden.txt</code> and
+            <code>name%20with%20spaces.txt</code> within the <code>v1/content</code> directory. This
```
ahankinson: Why are spaces and punctuation urlencoded?

ahankinson:
While I think we do need a rename example, I'm not sure this particular example gives a compelling case for where you might want to do that. 

Perhaps a more straightforward example, like "changing filename between versions" would be more straightforward?

draft/spec/index.html
awoods: My understanding is that "existing path" is now "content path".
https://github.com/OCFL/spec/issues/313

draft/spec/index.html
```diff
@@ -1031,6 +1031,43 @@ <h2>Versioned OCFL Object</h2>
...
+            specification does not constrain the normalization or renaming strategy adopted, in
+            this example URI encoding was used but many other approaches are possible, and they may
+            or may not yield an obvious relation between exiting file paths and logical file paths.
```
awoods: typo: "exiting" ..should likely be "content paths"

draft/spec/index.html
```diff
@@ -1031,6 +1031,43 @@ <h2>Versioned OCFL Object</h2>
...
+        </p>
+<pre>
+{
+  "digestAlgorithm": "sha512",
+  "head": "v1",
+  "id": "http://example.org/rename",
+  "manifest": {
+    "2deeaf...c0f": [ "v1/content/%2Ehidden.txt" ],
+    "329ebf...a6a": [ "v1/content/name%20with%20spaces.txt" ]
```
awoods: Are we assuming that certain OCFL clients may likely be implemented to URL-encode/decode the "content paths" when interacting with OCFL files on disk?

awoods:
> While I think we do need a rename example, I'm not sure this particular example gives a compelling case for where you might want to do that.
> 
> Perhaps a more straightforward example, like "changing filename between versions" would be more straightforward?

I was having similar thoughts. Or having a pairtree "content path" represented by a flat "logical path".

draft/spec/index.html
```diff
@@ -1031,6 +1031,43 @@ <h2>Versioned OCFL Object</h2>
 </pre>
     </section>
 
+    <section id="example-object-with-file-renaming">
+        <h2>File Renaming in an OCFL Object</h2>
```
rosy1280: Is this really file renaming?  Or is this more about file names

draft/spec/index.html
zimeon: I thought it best to stick with the current terminology until that is changed everywhere in response to #313

draft/spec/index.html
zimeon: Yes, I think based on feedback the term renaming title is wrong. I'll try (using the current/old terminology) "Different Existing File Paths and Logical File Paths in an OCFL Object" or (using the new terminology) "Different Storage Paths and Logical Paths in an OCFL Object".

draft/spec/index.html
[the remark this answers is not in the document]
zimeon: Fixed

draft/spec/index.html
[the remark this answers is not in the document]
zimeon: Fixed

draft/spec/index.html
zimeon: Based on comments I'll pick another example. I think a degree of filename normalization or sanitization will be normal but perhaps the URI encoding case is too confusing

draft/spec/index.html
zimeon: fixed "exiting" but planning to keep with old terminology pending fix for #313 (unless @rosy1280  gets to that before I get this PR approved ;-) )

zimeon:
Hmm... thinking about the two suggestions so far:
  * renaming between versions seems more complex than necessary because we can have paths that differ in just one version
  * a pairtree storage path doesn't make much sense unless the storage filename is also a hash (i.e. randomly distributed), in that case just using hashed filenames would be enough to show the example

so.. maybe the simplest example which gets the point across without focusing on the particular motivation is just to use a hash-based renaming?

zimeon:
OK my friends, have at it again... Remember that I am intentionally sticking with the current/old terminology of logical/existing files paths pending work on a PR for  #313

ahankinson:
It's better, and I have no *technical* reason why it shouldn't be used, so I've approved it.

However, it has caused me to think that this same mechanism might be used in anger, to facilitate filename obfuscation currently practiced in some software, e.g., SHA1 file renaming. I would be sad if our manifest / state mechanisms were used as a way of maintaining this practice, because it makes the actual digital content opaque to humans without the inventory file.

So I guess my only reservation with this example now is that I don't think it exemplifies "best practice" for making your content human-readable on-disk. I don't think we'll stop people from doing this, but I'm not entirely sure we should promote it in the spec either...

ahankinson:
@rosy1280 are you happy with the latest changes?

PR #339, opened 2019-04-18
Fixed: Clarify whitespace requirements in sidecar

### LINKED ISSUE #331: Digest sidecar file: whitespace separator
[So this seems really petty, but I'm trying to generate sample OCFL objects and inventory.json hand editing is painful.]

In the digest sidecar file, the OCFL spec requires a 'single space' between the digest and the string 'inventory.json'. However, sha512sum on Ubuntu generates output with a tab separator, not a single space.

In order to be strictly valid, I'm editing the output to convert the tab into a space. Would it be possible to accept a single whitespace character (e.g. space or tab) instead of requiring a space?

ahankinson:
This PR clarifies the requirement that one or more whitespace characters must separate the digest from the filename in the inventory sidecar file.

Fixes #331

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -719,7 +719,7 @@
 DIGEST inventory.json
 </pre>
         <p>
-            One or more linear whitespace characters (spaces or tabs) must separate DIGEST from the string
+            One or more whitespace characters (spaces or tabs) must separate DIGEST from the string
             <code>inventory.json</code>; that is, the name of the inventory file in the same directory.
         </p>
         <p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -713,9 +713,14 @@
             the chosen digest algorithm for the object. An example might be <code>inventory.json.sha512</code>.
         </p>
         <p>
-            The digest sidecar file MUST contain the digest of the inventory file. This MUST follow the format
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -713,9 +713,14 @@ <h2>Inventory Digest</h2>
...
-            The digest sidecar file MUST contain the digest of the inventory file. This MUST follow the format
-            <code>DIGEST inventory.json</code>; that is, the digest of the inventory file, a single space, and the
-            string <code>inventory.json</code> which is the name of the inventory file in the same directory.
+            The digest sidecar file MUST contain the digest of the inventory file. This MUST follow the format:
+        </p>
+<pre>
+DIGEST inventory.json
+</pre>
+        <p>
+            One or more linear whitespace characters (spaces or tabs) must separate DIGEST from the string
```
zimeon: I think "linear whitespace" is rather confusingly defined in https://www.rfc-editor.org/std/std68.txt (in one place as space+tab, in another including CRLF; see https://stackoverflow.com/questions/21072713/what-exactly-is-the-linear-whitespace-lws-lwsp) and in https://www.ietf.org/rfc/rfc2616.txt (`LWS            = [CRLF] 1*( SP | HT )`) so I suggest removing the word "linear"

PR #342, opened 2019-04-19
Proposal to address #341

### LINKED ISSUE #341: Relax requirement for 'content' directory
Moab requires each version directory to contain a 'data' and a 'manifests' directory, which makes an existing Moab object non-compliant with 0.2 and unable to become so.

I propose we relax the spec to enable a version directory to contain any number of directories of any name, but only able to contain two files in the root - inventory.json and inventory.json.sha512. This enables existing Moab objects to become OCFL-compliant by adding the inventory.json and sidecar.

julianmorley:
Fixes #341

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -393,17 +393,15 @@
             version directory names.
         </p>
         <p>
-            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
-            SHOULD NOT contain a <code>content</code> sub-directory otherwise. There MUST be no other files or
-            directories as children of a version directory, other than an <a href="#inventory">inventory file</a>,
-            an <a href="#inventory-digest">inventory digest</a>, and a <code>content</code> directory.
+            Version directories MAY contain one or more sub-directories. These sub-directories may have any valid name.
+            There MUST be no files as children of a version directory, other than an <a href="#inventory">inventory file</a>,
+            and an <a href="#inventory-digest">inventory digest</a>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


zimeon:
This PR removes the suggestion that the sub-directories SHOULD NOT be present and empty (which we had for `content`). Not sure whether intentional

ahankinson:
I would still like to see a recommendation for a `content` directory. The point of moving the version content into a sub-directory was to ensure we did not mix the administrative workings of OCFL with the content being managed. 

This change is introduced out of a desire for backwards compatibility with Moab which is desirable, but perhaps we can tighten this up to say they should only be 'content', or 'data' and 'manifests' for backwards-compatibility. 

Opening it up to arbitrary directories may otherwise risk users misinterpreting the relaxing of this rule.

draft/spec/index.html
```diff
@@ -393,17 +393,15 @@ <h2>Version Directories</h2>
             version directory names.
         </p>
         <p>
-            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
-            SHOULD NOT contain a <code>content</code> sub-directory otherwise. There MUST be no other files or
-            directories as children of a version directory, other than an <a href="#inventory">inventory file</a>,
-            an <a href="#inventory-digest">inventory digest</a>, and a <code>content</code> directory.
+            Version directories MAY contain one or more sub-directories. These sub-directories may have any valid name.
```
awoods: What constitutes a "valid name"?

PR #346, opened 2019-05-01
Change logical and existing path terms in implementation notes

### LINKED ISSUE #313: Consider changing existing file path to something else

zimeon:
Closes #313 (in combination with #333 which is already merged)

Turns out that we don't use the formal terms much in the implementation notes


rosy1280:
Thanks @zimeon

PR #349, opened 2019-05-15
Add "contentDirectory" to support non-standard content dir names.

julianmorley:
Satisfies #341

### DIFF
--- a/0.2/spec/index.html
+++ b/0.2/spec/index.html
@@ -410,7 +410,7 @@
         </p>
         <p>
             If the key <code>contentDirectory</code> is not present in the <a href="#inventory">inventory file</a> then the
-            name of the designated content sub-directory MUST be <code>content</content>. OCFL-compliant tools
+            name of the designated content sub-directory MUST be <code>content</code>. OCFL-compliant tools
             (including any validators) MUST ignore all directories in the object version directory except for the
             designated content directory.
         </p>
--- a/0.2/spec/index.html
+++ b/0.2/spec/index.html
@@ -396,24 +396,23 @@
         <p>
             Version directories MUST contain a designated content sub-directory if the object contains files to be
             preserved, and SHOULD NOT contain this sub-directory otherwise. The name of this designated sub-directory
-            MAY be defined in the <a href="#inventory">inventory file</a> using the key "contentDirectory" with the
-            value being the chosen sub-directory name as a string, relative to the version directory. The value of the
-            key "contentDirectory" MUST NOT contain any filesystem delimiter characters. If the key "contentDirectory"
-            is set, it MUST be set in the first version of the object and MUST NOT change between versions of the
-            same object.
+            MAY be defined in the <a href="#inventory">inventory file</a> using the key <code>contentDirectory</code>
+            with the value being the chosen sub-directory name as a string, relative to the version directory.
+            The value of the key <code>contentDirectory</code> MUST NOT contain any filesystem delimiter characters.
+            If the key <code>contentDirectory</code> is set, it MUST be set in the first version of the object and MUST NOT
+            change between versions of the same object.
           </p>
           <p>
             There MUST be no other files as children of a version directory, other than an <a
              href="#inventory">inventory file</a> and a <a href="#inventory-digest">inventory digest</a>.
-            The version directory SHOULD NOT contain any directories other than the designated <code>content</code>
+            The version directory SHOULD NOT contain any directories other than the designated content
             sub-directory.
         </p>
         <p>
-            If the key "contentDirectory" is not present in the <a href="#inventory">inventory file</a> then the name
-            of the designated content sub-directory MUST be 'content'. OCFL-compliant tools (including any validators)
-            MUST ignore all directories in the object root except for the designated content directory. OCFL
-            validators may log the presence of directories that are not the designated content directory, but must
-            not fail validation for that reason.
+            If the key <code>contentDirectory</code> is not present in the <a href="#inventory">inventory file</a> then the
+            name of the designated content sub-directory MUST be <code>content</content>. OCFL-compliant tools
+            (including any validators) MUST ignore all directories in the object version directory except for the
+            designated content directory.
         </p>
         <p>
             Every file within a version's <code>content</code> directory MUST be referenced in the
--- a/0.2/spec/index.html
+++ b/0.2/spec/index.html
@@ -394,10 +394,26 @@
             version directory names.
         </p>
         <p>
-            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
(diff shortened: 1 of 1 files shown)


0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            MAY be defined in the <a href="#inventory">inventory file</a> using the key "contentDirectory" with the
+            value being the chosen sub-directory name as a string, relative to the version directory. The value of the
+            key "contentDirectory" MUST NOT contain any filesystem delimiter characters. If the key "contentDirectory"
+            is set, it MUST be set in the first version of the object and MUST NOT change between versions of the
+            same object.
+          </p>
+          <p>
+            There MUST be no other files as children of a version directory, other than an <a
+             href="#inventory">inventory file</a> and a <a href="#inventory-digest">inventory digest</a>.
+            The version directory SHOULD NOT contain any directories other than the designated <code>content</code>
```
awoods: Potentially remove the `code` element to be consistent with the text in the first sentence.

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            sub-directory.
+        </p>
+        <p>
+            If the key "contentDirectory" is not present in the <a href="#inventory">inventory file</a> then the name
+            of the designated content sub-directory MUST be 'content'. OCFL-compliant tools (including any validators)
+            MUST ignore all directories in the object root except for the designated content directory. OCFL
```
awoods: Allowing random directories in `version` directories infringes on the purity of OCFL Objects, and therefore makes me a bit concerned.
I assume this is the loophole through which legacy Moab Objects will be migrated?

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
             version directory names.
         </p>
         <p>
-            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
-            SHOULD NOT contain a <code>content</code> sub-directory otherwise. There MUST be no other files or
-            directories as children of a version directory, other than an <a href="#inventory">inventory file</a>,
-            an <a href="#inventory-digest">inventory digest</a>, and a <code>content</code> directory.
+            Version directories MUST contain a designated content sub-directory if the object contains files to be
+            preserved, and SHOULD NOT contain this sub-directory otherwise. The name of this designated sub-directory
+            MAY be defined in the <a href="#inventory">inventory file</a> using the key "contentDirectory" with the
```
zimeon: "contentDirectory" -> \<code>contentDirectory\</code>

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            value being the chosen sub-directory name as a string, relative to the version directory. The value of the
+            key "contentDirectory" MUST NOT contain any filesystem delimiter characters. If the key "contentDirectory"
```
zimeon: More \<code>contentDirectory\</code> x 2

0.2/spec/index.html
zimeon: +1

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            same object.
+          </p>
+          <p>
+            There MUST be no other files as children of a version directory, other than an <a
+             href="#inventory">inventory file</a> and a <a href="#inventory-digest">inventory digest</a>.
+            The version directory SHOULD NOT contain any directories other than the designated <code>content</code>
+            sub-directory.
+        </p>
+        <p>
+            If the key "contentDirectory" is not present in the <a href="#inventory">inventory file</a> then the name
```
zimeon: "contentDirectory" -> \<code>contentDirectory\</code>

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            of the designated content sub-directory MUST be 'content'. OCFL-compliant tools (including any validators)
```
zimeon: 'content' -> \<code>content\</code>

0.2/spec/index.html
zimeon: s/in the object root/in the version directories/

I share some concern about this loophole but unless we opt for allowing multiple content directories (which I think the consensus on #341 was against) then I don't see a better option for legacy support than this.

0.2/spec/index.html
```diff
@@ -394,10 +394,26 @@ <h2>Version Directories</h2>
...
+            MUST ignore all directories in the object root except for the designated content directory. OCFL
+            validators may log the presence of directories that are not the designated content directory, but must
```
zimeon: We have non-normative style `may` and `must not` in this sentence. I think we don't need the last part of this sentence (_", but not fail validation for that reason."_) because that simply duplicates the preceding sentence. I'd be kinda tempted to skip the whole sentence because we haven't defined what it means for a validator to log something.

0.2/spec/index.html
zimeon: Do we need to be more specific about what a "filesystem delimeter character" is?

julianmorley: Resolved the contentDirectory changes; I'm unsure what to do about the delimiter issue. Essentially trying to prevent someone from defining contentDir as 'my/content'. Maybe restrict the values to alphanumerics?

0.2/spec/index.html
julianmorley: Removed the whole sentence.

0.2/spec/index.html
julianmorley: We could instead just conform to DNS hostname conventions; 'Valid characters for hostnames are ASCII(7) letters from a to z, the digits from 0 to 9, and the hyphen (-).  A hostname may not start with a hyphen.' Which is a bit limiting but certainly avoids the file delimiter issue and any potential filecase issues.

zimeon: I'll admit to struggling with this one. A restriction such as DNS hostname conventions seems intuitively reasonable but also arbitrary. Why would it be wrong to use a word in some language and character set other than English and ASCII? We don't put constraints like this anywhere else.

ahankinson: Hm. I think that would cause more confusion than it solves.

https://i❤.ws/emojidomain/💩 is a valid DNS entry...

zimeon: I think we should says simply

> The `contentDirectory` value MUST NOT contain the forward slash (`/`) path separator.

This is consistent with our description of [Inventory](https://ocfl.io/draft/spec/#inventory) where we define the special meaning as a delimiter.

PR #353, opened 2019-05-21
Add contentDirectory to schema

birkland:
* Adds `contentDirectory` as an allowable property to the inventory.json schema
* Restricts the value of `contentDirectory` to any string that does not contain a solidus (`/`) character

One oddity:  This PR is targeted towards the `draft/` directory, though the text defining `contentDirectory` seems to be in `0.2/` directory, which I _think_ was supposed to be immutable after its publication.  Let me know if this PR should be to `0.2/` instead.

### DIFF
--- a/draft/spec/inventory_schema.json
+++ b/draft/spec/inventory_schema.json
@@ -71,6 +71,10 @@
                     "required": [ "created", "state" ]
                 }
             }
+        },
+        "contentDirectory": {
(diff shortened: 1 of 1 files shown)


zimeon:
Thanks @birkland. PR on `/draft/` is what we want, thanks for catching the other error

zimeon:
Also connected with #341

PR #355, opened 2019-05-21
Reapply contentdir_patch against draft spec

### LINKED ISSUE #341: Relax requirement for 'content' directory

zimeon:
Same as #349 but applied to `/draft/` spec instead of old `/0.2/` versions, really fixes #341 this time ;-)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -400,10 +400,25 @@
             version directory names.
         </p>
         <p>
-            Version directories MUST have a sub-directory called <code>content</code> if there are files present, and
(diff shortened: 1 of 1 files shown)


zimeon:
And, of course, the old PR never got the line length check because it was on the wrong file. So now I have fixed that too

Issue #352, opened 2019-05-21
Update Moab example in spec to reflect contentDirectory usage


julianmorley:
Moab would use "contentDirectory": "data"; update example as appropriate.

rosy1280:
closed by pull request #356

PR #356, opened 2019-05-21
Update Moab example to reflect contentDirectory changes

julianmorley:
Addresses #352

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -1455,6 +1455,6 @@ myfirstbag
   }
 </pre>
 </section>
-
+</section>
 </body>
 </html>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -1278,9 +1278,9 @@ myfirstbag
             OCFL conversion tool. This tool uses the Moab-versioning gem to extract deltas and digests
             of the Moab data directory for each Moab version and translate those into version state blocks
             in an OCFL inventory file, which would be placed in the root directory of the Moab object.
-            The content of the /data directory in the Moab version directories (and thus, the bitstreams
+            The content of the <code>data</code> directory in the Moab version directories (and thus, the bitstreams
             that Moab is preserving) is tracked by OCFL, via the <code>contentDirectory</code> value.
-            The contents of the Moab /manifests directories are not tracked, as the intention is not
+            The contents of the Moab <code>manifests</code> directories are not tracked, as the intention is not
             to encapsulate a Moab object inside an OCFL object, but rather to migrate Moab's preserved
             bitstreams into an OCFL object without compromising legacy access patterns.
         </p>
@@ -1356,9 +1356,10 @@ myfirstbag
                   └── versionInventory.xml
 </pre>
         <p>
-            An OCFL inventory that tracks the /data directory would include a manifest comprised as follows.
-            Note the absence of the /manifests directory, as we are not encapsulating the Moab object in an OCFL object,
-            and the presence of <code>contentDirectory</code> to specify /data as the preserved content directory:
+            An OCFL inventory that tracks the <code>data</code> directory would include a manifest comprised as follows.
+            Note the absence of the <code>manifests</code> directory, as we are not encapsulating the Moab object
+            in an OCFL object, and the presence of <code>contentDirectory</code> to specify <code>data</code> as the
+            preserved content directory:
         </p>
 <pre>
   {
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -1278,9 +1278,11 @@ <h2>Moab in an OCFL Object</h2>
             OCFL conversion tool. This tool uses the Moab-versioning gem to extract deltas and digests
             of the Moab data directory for each Moab version and translate those into version state blocks
             in an OCFL inventory file, which would be placed in the root directory of the Moab object.
-            All extant files in the Moab version directories are tracked in the manifest block. The contents
-            of the Moab manifests directory are not tracked in the version state blocks; they are effectively
-            deleted from the OCFL object without compromising legacy access patterns.
+            The content of the /data directory in the Moab version directories (and thus, the bitstreams
```
zimeon: Change `cand I think we should drop the leading slash -- I'm really not sure what that means given that one already says that `data` is in the version directory

draft/spec/index.html
```diff
@@ -1278,9 +1278,11 @@ <h2>Moab in an OCFL Object</h2>
...
+            that Moab is preserving) is tracked by OCFL, via the <code>contentDirectory</code> value.
+            The contents of the Moab /manifests directories are not tracked, as the intention is not
```
zimeon: `/manifests` -> `<code>manifests</code>`

draft/spec/index.html
```diff
@@ -1296,172 +1298,162 @@ <h2>Moab in an OCFL Object</h2>
...
+              └── manifests
+                  ├── fileInventoryDifference.xml
+                  ├── manifestInventory.xml
+                  ├── signatureCatalog.xml
+                  ├── versionAdditions.xml
+                  └── versionInventory.xml
 </pre>
         <p>
-            An OCFL inventory that tracks the /data directory would include a manifest comprised as follows:
+            An OCFL inventory that tracks the /data directory would include a manifest comprised as follows.
```
zimeon: `/data` --> `<code>data</code>`

draft/spec/index.html
```diff
@@ -1296,172 +1298,162 @@ <h2>Moab in an OCFL Object</h2>
...
+            Note the absence of the /manifests directory, as we are not encapsulating the Moab object in an OCFL object,
+            and the presence of <code>contentDirectory</code> to specify /data as the preserved content directory:
```
zimeon: `/data` --> `<code>data</code>`

draft/spec/index.html
```diff
@@ -1296,172 +1298,162 @@ <h2>Moab in an OCFL Object</h2>
...
+                  ├── fileInventoryDifference.xml
+                  ├── manifestInventory.xml
+                  ├── signatureCatalog.xml
+                  ├── versionAdditions.xml
+                  └── versionInventory.xml
 </pre>
         <p>
-            An OCFL inventory that tracks the /data directory would include a manifest comprised as follows:
+            An OCFL inventory that tracks the /data directory would include a manifest comprised as follows.
+            Note the absence of the /manifests directory, as we are not encapsulating the Moab object in an OCFL object,
```
zimeon: `/manifests` --> `<code>manifests</code>`

draft/spec/index.html
```diff
@@ -1296,172 +1298,162 @@ <h2>Moab in an OCFL Object</h2>
...
+        "509a2d...dc6": [ "v0003/data/metadata/identityMetadata.xml" ],
+        "548066...893": [ "v0003/data/metadata/versionMetadata.xml" ],
+        "93884e...aae": [ "v0003/data/metadata/contentMetadata.xml" ],
+        "4c5ab4...b02": [ "v0003/data/metadata/embargoMetadata.xml" ]
+    }
+    ...
 </pre>
     <p>
-        The version 1 state block would look like this. Note the absence of the manifests/ directory.
+        The version 1 state block would look like this.
```
zimeon: Instead of separating out the `manifest` and three `state` blocks into separate code examples, would it be better to simply write out the complete inventory as one example? There would then be no doubt about what else should or should not be present _AND_ we could validate it.

draft/spec/index.html
julianmorley: Resolved in new commit

draft/spec/index.html
julianmorley: Resolved in new commit

draft/spec/index.html
julianmorley: Resolved in new commit

draft/spec/index.html
julianmorley: Resolved in new commit

draft/spec/index.html
julianmorley: Consolidated example into a single inventory file ; it's hand-rolled so I hope it validates.

Issue #320, opened 2019-03-14
Add to implementation notes a discussion of the idea of temporary space while building OCFL Objects


zimeon:
From discussion in [2019.03.13 Community Meeting](https://github.com/OCFL/spec/wiki/2019.03.13-Community-Meeting) there might be a need for a "draft" or "tmp" directory for active  OCFL Objects.

zimeon:
I wonder whether there might be a case for a normative (specification) addition that says some particular directory (maybe `tmp`) _MAY_ be present but _MUST_ be ignored. This would support validation of an object that is actively being worked on for all but the final creation of the new version (something along the lines of `mv tmp vN; cp -p vN/inventory.json vN/inventory.json.sha512 .`).

awoods:
Sounds like a similar description for the current `logs` directory.

birkland:
I toyed with the idea of placing the in-progress directory somewhere  under the `logs` directory (e.g. `logs/.v3`), vs simply building up a `content/v3` that is unreferenced in any inventory files until the very end when the work is "committed".  I ended up doing the latter, justifying it by the notion that there is always a period of time when the object is in an inconsistent state when copying it into place, or adding a new version - so why not embrace it.  The purview of the spec is objects at rest anyway, and building up a version is motion.

A temp directory (or some sort of convention, like directories under content that begin with a dot `.` MUST be ignored) does seem cleaner, though.

ahankinson:
What about `.vX` (in-progress) vs `vX` (committed)? So `.v3` while the content is being assembled, and then moved to `v3` when it's finished.

zimeon:
Something along the lines of `vX` would be fine but I don't see any advantage to hiding with a leading dot

julianmorley:
A temporary space on the storage root, *not* named /tmp, for the assembly of future OCFL object versions.

rosy1280:
we agreed to use the language in @zimeon second comment, except the directory should be named deposit and be placed at the storage root.

zimeon:
Hmmm... as @rosy1280 just pointed out on my abortive PR #324, the https://github.com/OCFL/spec/issues/320#issuecomment-474907680 above suggests one `deposit` directory on the storage root. I'm not opposed to such a directory which might be useful for the assembly of **new objects**, but I do not think it is a good solution for the issue debated on this ticket, which is primarily **updating objects with new versions**.

[...]

PR #362, opened 2019-06-04
Guidance on versioning implementation 1

neilsjefferies:
This should work regardless of where the temp files are locate. There needs to be some temp files when we write to the object inventory anyway.

### DIFF
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -629,7 +629,7 @@
           <li>  
           Temporary Version Directories: New versions should be created in a temporary version directory.
           This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
-          other version directories in an OCFL Object or elsewhere. in any case, only one such directory 
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory 
           can exist per OCFL Object at any time and updates to it should be managed by a single 
           controlling process to avoid conflicts. It is advised that object updates should be given
           a unique transaction identifier to simplify any overlying process control logic. For 
@@ -687,13 +687,13 @@
             </li>
             <li>
             Does a temporary version directory for the object and transaction ID in question exist?
-	    If no, then abort with error. transaction ID is invalid for some reason, need to debug!
+	    If no, then abort with error. The transaction ID is invalid for some reason, need to debug!
             </li>
             <li>
-            Update object as per defined operations section earlier
+            Update object as described earlier.
             </li>
             <li>
-            Ideally, update local inventory after each operation, not at the end
+            Ideally, update the local inventory after each file change operation, not at the end
             </li>
           </ol>
         </section>
@@ -706,7 +706,7 @@
             </li>
             <li>
             Does a temporary version directory for the object and transaction ID in question exist?
-	    If no, then abort with error. transaction ID is invalid for some reason, need to debug!
+	    If no, then abort with error. The transaction ID is invalid for some reason, need to debug!
             </li>
             <li>
             Generate inventory checksum file in temporary version directory
@@ -738,16 +738,18 @@
             Delete temporary version directories and inventory files - this automatically reverts to last good version.
             </li>
             <li>
-            If the manifest checksum fails then you will find a new version directory not in the manifest
+            If the manifest checksum fails then the manifest is corrupted by a failed copy. This should be recoverable
+            from the most recent version directory (which is the newly created one).
             </li>
             <li>
-            Any overlying transactional store will need cleanup but basically repeat 5-7 above
+            Any overlying transactional store will need cleanup but basically repeat 5-7 above after validating all
+            object checksums - considering some sort of failure has just occurred.
             </li>    
           </ol>
         </section>
 
       </section>
-
+    </section>
   </section>
 
 </body>
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -602,6 +602,151 @@
         </li>
       </ul>
     </section>
+    
+    <section>
+      <h2>Guidance on Updating OCFL Object Versions</h2>
+      <p>
+      The OCFL is designed to be a specification that covers objects at rest and consequently
+      does not specify in detail update and file locking mechanisms since these are implementation
+      dependent features. Nevertheless, this section includes some basic guidance on how to 
+      update OCFL objects in a manner that tries to ensure that updates are limited to a single  
+      transaction and that failures are detectable and recoverable.
+      </p>
+    
+      <section>
+        <h2>Temporary Names</h2>
+        <p>  
+        While an OCFL object is being updated, and a new version is created, files and folders 
+        being written to should have temporary names that are not meaningful to OCFL parsers.
+        The top level inventory should also be updated last. This ensures that all the files 
+        referenced by the inventory are valid and, consequently, read-only clients that
+        reference the inventory should continue to operate normally except for the brief moment
+        when the inventory is actually being updated. In practice, it is expected that upstream
+        web caching should be able to cover this momentary unavailability in the majority of 
+        cases. The following temporary names are defined:
+        <ul>
+          <li>  
+          Temporary Version Directories: New versions should be created in a temporary version directory.
+          This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
+          other version directories in an OCFL Object or elsewhere. in any case, only one such directory 
+          can exist per OCFL Object at any time and updates to it should be managed by a single 
+          controlling process to avoid conflicts. It is advised that object updates should be given
+          a unique transaction identifier to simplify any overlying process control logic. For 
+          example, the identifier could be used as an an eTag for a REST API implementation. A 
+          good temporary directory name should be "deposit_transaction ID_OCFL Object ID". 
+          If there is guaranteed to only be a single process controlling updates to the contents 
+          of an OCFL storage root, then it is permissible to use "deposit_OCFL Object ID" as a 
+          temporary version directory name. If temporary directories are stored within OCFL 
+          Objects then the OCFL Object ID part may be omitted since it is implicit.
+          </li>
+          <li>
+          Temporary Inventory Files: A top level object inventory file should never be updated in situ.
+          Instead, the inventory should be created within the temporary version directory and updated as
+          the version contents are modified. This provides some level of forensic recovery information in 
+          the event of failure during version creation, yet ensures that the top level inventory remains 
+          valid for read clients for as long as possible. When the construction of the new version 
+          and its inventory is complete, it can then be copied to a temporary file name alongside 
+          the top level inventory. Temporary inventory files always occur in OCFL Object Roots and
+          should be names "tmp_inventory.json".
+          </li>
+        </ul>
+      </section>
+
+      <section>
+      <h2>Operational Logic</h2>
+      <p>
+      Bearing in mind the temporary names specified above, it is possible to sketch out how  
+      they can be used to ensure some level of integrity in during OCFL updates.
+      </p>
+
+        <section>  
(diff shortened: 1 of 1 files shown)


draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+        reference the inventory should continue to operate normally except for the brief moment
+        when the inventory is actually being updated. In practice, it is expected that upstream
+        web caching should be able to cover this momentary unavailability in the majority of 
+        cases. The following temporary names are defined:
+	</p>	
+        <ul>
+          <li>  
+          Temporary Version Directories: New versions should be created in a temporary version directory.
+          This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory
```
rosy1280: what the does "the same location as other version directories in an OCFL Object or elsewhere" mean.

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+        when the inventory is actually being updated. In practice, it is expected that upstream
+        web caching should be able to cover this momentary unavailability in the majority of 
+        cases. The following temporary names are defined:
+	</p>	
+        <ul>
+          <li>  
+          Temporary Version Directories: New versions should be created in a temporary version directory.
+          This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory 
+          can exist per OCFL Object at any time and updates to it should be managed by a single
```
rosy1280: only one deposit directory can exist per object?  only one temporary version directory can exist per object? this isn't clear.

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+        cases. The following temporary names are defined:
+	</p>	
+        <ul>
+          <li>  
+          Temporary Version Directories: New versions should be created in a temporary version directory.
+          This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory 
+          can exist per OCFL Object at any time and updates to it should be managed by a single 
+          controlling process to avoid conflicts. It is advised that object updates should be given
+          a unique transaction identifier to simplify any overlying process control logic. For
```
rosy1280: This is way to into the weeds on how to architect a repository.

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+          Temporary Version Directories: New versions should be created in a temporary version directory.
+          This could reside in the OCFL Storage Root workspace directory ("deposit"), the same location as
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory 
+          can exist per OCFL Object at any time and updates to it should be managed by a single 
+          controlling process to avoid conflicts. It is advised that object updates should be given
+          a unique transaction identifier to simplify any overlying process control logic. For 
+          example, the identifier could be used as an an eTag for a REST API implementation. A 
+          good temporary directory name should be "deposit_transaction ID_OCFL Object ID". 
+          If there is guaranteed to only be a single process controlling updates to the contents 
+          of an OCFL storage root, then it is permissible to use "deposit_OCFL Object ID" as a
```
rosy1280: If I have a top level deposit directory, then I don't need to name a directory this because I know everything in that directory is for deposit into the storage root.

draft/implementation-notes/index.html
neilsjefferies: I'm not saying where the workspace is....just that there has to be one and the contents are predictable and therefore identifiable.

draft/implementation-notes/index.html
neilsjefferies: You're right. The temporary version directory is called "deposit..."

draft/implementation-notes/index.html
neilsjefferies: There might be multiple transactions going on in that directory so they need separate distinguishable directories. Embedding a transaction ID also gives protection against multiple writes to the same object.

draft/implementation-notes/index.html
neilsjefferies: Bingo! I refer you to Fedora 4 and DPN - apparently this is something some people find rather hard. This is just a guideline on how not to screw it up too badly.

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+          other version directories in an OCFL Object or elsewhere. In any case, only one such directory 
+          can exist per OCFL Object at any time and updates to it should be managed by a single 
+          controlling process to avoid conflicts. It is advised that object updates should be given
+          a unique transaction identifier to simplify any overlying process control logic. For 
+          example, the identifier could be used as an an eTag for a REST API implementation. A 
+          good temporary directory name should be "deposit_transaction ID_OCFL Object ID". 
+          If there is guaranteed to only be a single process controlling updates to the contents 
+          of an OCFL storage root, then it is permissible to use "deposit_OCFL Object ID" as a 
+          temporary version directory name. If temporary directories are stored within OCFL 
+          Objects then the OCFL Object ID part may be omitted since it is implicit.
```
rosy1280: so you're saying that there would be just a `V2` directory? how would you know that a version hasn't been completed in the directory?

draft/implementation-notes/index.html
rosy1280: that i would change the sentence to say this, because right now i don't really know what that sentence says.

draft/implementation-notes/index.html
rosy1280: i'm suggesting that you remove this from the language because it shouldn't be on us to tell folks how to architect their repository.

draft/implementation-notes/index.html
rosy1280: that's one way to do it, but another is to just not let people write to the same object while another version is being deposited.  see comment above about this being way too into the weeds and we shouldn't be telling folks what to do.

rosy1280:
@neilsjefferies I haven't finished reviewing this, but I would suggest that we re-visit the previous conversations we had around deposit directories and identify what this section's purpose is because right now this doesn't necessarily seem like implementation guidance.

draft/implementation-notes/index.html
zimeon: I seems really hard to discuss a semi-blessed `deposit` directory inside objects in a PR. May I suggest creating a new issue (that points back to #320) that proposes guidance on this issue in the implementation notes while not changing the normative specification?

Note also should use code font `<code>deposit</code>` instead of quoting literals

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+      <p>
+      The OCFL is designed to be a specification that covers objects at rest and consequently
+      does not specify in detail update and file locking mechanisms since these are implementation
+      dependent features. Nevertheless, this section includes some basic guidance on how to 
+      update OCFL objects in a manner that tries to ensure that updates are limited to a single  
+      transaction and that failures are detectable and recoverable.
+      </p>
+    
+      <section>
+        <h2>Temporary Names</h2>
```
zimeon: "Temporary Names" sounds odd to me, should this be "Temporary Files and Directories"?

draft/implementation-notes/index.html
zimeon: I wouldn't recommend directory names with spaces but maybe that is just old-school of me.

I think we also have to be careful around language like _permissible_ as this is non-normative, the issue here is that with just a single process it is _adequate_ without the transaction id.

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
     </section>
+    
+    <section>
+      <h2>Guidance on Updating OCFL Object Versions</h2>
+      <p>
+      The OCFL is designed to be a specification that covers objects at rest and consequently
+      does not specify in detail update and file locking mechanisms since these are implementation
+      dependent features. Nevertheless, this section includes some basic guidance on how to 
+      update OCFL objects in a manner that tries to ensure that updates are limited to a single  
+      transaction and that failures are detectable and recoverable.
```
zimeon: Do we need to point out that advice here will result in objects that cannot be validated during update, validation should be performed only when updates are complete?

draft/implementation-notes/index.html
```diff
@@ -602,7 +602,154 @@ <h2>Forward Delta</h2>
...
+            </li>
+            <li>
+            Does a temporary version directory for the object and transaction ID in question exist?
+	    If no, then abort with error. The transaction ID is invalid for some reason, need to debug!
+            </li>
+            <li>
+            Update object as described earlier.
+            </li>
+            <li>
+            Ideally, update the local inventory after each file change operation, not at the end
```
zimeon: Not sure whether this is ideal. If one thinks of an atomic update of many files then it is just a waste of time

draft/implementation-notes/index.html
neilsjefferies: Yes, shouldn't be spaces - will fix that. How often will we ever have a single process controlling access since it is likely that any overlying API will be RESTful. 

This is in the implementation notes so not part of the spec and therefore shouldn't it all be non-normative. Adequate is adequate however!

draft/implementation-notes/index.html
neilsjefferies: No, but in any case, version completeness can be determined from the inventory.

draft/implementation-notes/index.html
neilsjefferies: That's the point - the only way to do any form of locking is to stick a flag somewhere. Since OCFL  aims to do away with non-file-based data, this is an OCFL-ish way to do it. This is guidance, as stated above, not the spec so it's just a suggestion.

draft/implementation-notes/index.html
neilsjefferies: ...this *is* guidance in the implementation notes

zimeon: Sorry for not being clear: my point is that it would be good to merge other parts of this PR but I don't think we can introduce a new in-object `deposit` directory without more discussion on a separate issue

draft/implementation-notes/index.html
neilsjefferies: Agreed...done

draft/implementation-notes/index.html
neilsjefferies: Noted.

draft/implementation-notes/index.html
neilsjefferies: Limited discussion to "deposit" directory in Storage Root.

draft/implementation-notes/index.html
neilsjefferies: Hoopefully clarified.

draft/implementation-notes/index.html
neilsjefferies: Scrapped.

draft/implementation-notes/index.html
neilsjefferies: Scrapped.

neilsjefferies: Mostly scrapped.

draft/implementation-notes/index.html
neilsjefferies: Scrapped.

draft/implementation-notes/index.html
neilsjefferies: Agreed. Advantages both ways depending on use case. Scrapped as we don;t really need to say anything about it.

PR #378, opened 2019-10-02
Require mapping from id to storage path (#374)

### LINKED ISSUE #374: Make it clear that it is expected to have a well defined bi-directional mapping between storage path and id
The current text in [4.3 Storage Hierarchies](https://ocfl.io/0.3/spec/#root-hierarchies) does not spell out that we expect there to be a well-defined bi-directional mapping between id and storage paths such that for an id the path is uniquely defined, and vice versa. 

See https://github.com/OCFL/spec/issues/369#issuecomment-531680861 and following discussion

zimeon:
Adds clarification, fixes #374

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -913,7 +913,7 @@ DIGEST inventory.json
           hierarchies, the following restrictions apply:
         </p>
         <ol>
-          <li>There MUST be a well-defined mapping from an object identifier to a unique storage path</li>
+          <li>There MUST be a deterministic mapping from an object identifier to a unique storage path</li>
           <li>Storage hierarchies MUST NOT include files within intermediate directories</li>
           <li>Storage hierarchies MUST be terminated by OCFL Object Roots</li>
           <li>Storage hierarchies within the same OCFL Storage Root SHOULD use just one layout pattern</li>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -913,12 +913,12 @@ <h2>Storage Hierarchies</h2>
           hierarchies, the following restrictions apply:
         </p>
         <ol>
+          <li>There MUST be a well-defined mapping from an object identifier to a unique storage path</li>
```
awoods:
```
There MUST be a well-defined, [reproducible] mapping...
```

zimeon: Do you really want to add "reproducible"? Seems over the top to me

rosy1280: yeah, given feedback so far, i would agree that adding reproducible might be a bit much.

ahankinson: Perhaps "deterministic"?

zimeon: well-defined is deterministic ... see https://en.wikipedia.org/wiki/Well-defined

zimeon: agree with discussion to change from well-defined to deterministic

PR #381, opened 2019-10-02
Updates Object Structure Section

### LINKED ISSUE #373: Clarify whether and OCFL Object Root may contain other files/directories not specified
By my reading of [3. OCFL Object](https://ocfl.io/0.3/spec/#object-spec) it is not clear whether  files or directories not specified in section 3 (conformance decl., version dirs, inventory, inventory checksum, logs dir) are permitted. I think we should add language to [3.1 Object Structure](https://ocfl.io/0.3/spec/#object-structure) along the lines of either:

> The OCFL Object Root MUST NOT contain files or directories other than those specified in the following sections.

or:

> The OCFL Object Root MAY contain additional files or directories not specified in the following sections. OCFL-compliant tools (including any validators) MUST ignore and such additional files or directories.

(or one could allow just files, or just directories)

rosy1280:
Fixes #373

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -385,10 +385,10 @@
         └── content
                └── ... content files ...
 </pre>
-            <p>
-              The <a>OCFL Object Root</a> MUST NOT contain files or directories other than those specified in the
-              following sections.
-            </p>
+        <p>
+            The <a>OCFL Object Root</a> MUST NOT contain files or directories other than those specified in the
+            following sections.
+        </p>
     </section>
 
     <section id="object-conformance-declaration">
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -385,17 +385,18 @@ <h2>Object Structure</h2>
         └── content
                └── ... content files ...
 </pre>
+            The <a>OCFL Object Root</a> MUST NOT contain files or directories other than those specified in the
+            following sections.
```
ahankinson: Needs wrapping in `<p>` tags

zimeon: Good catch!

PR #382, opened 2019-10-02
Expand on when it is OK to not have an inventory in a version directory.

julianmorley:
This is to close the loophole where it is apparently OK to delete inventory files in prior versions, an unintentional consequence of the wording chosen to support migration from previous systems with immutable version directories.

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -794,12 +794,14 @@ DIGEST inventory.json
             SHOULD include an inventory file that is an <a href="#inventory">Inventory</a> of all content for versions
             up to and including that particular version. Where an OCFL Object contains <code>inventory.json</code>
             in version directories, the inventory file in the OCFL Object Root MUST be the same as the file in the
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -794,12 +794,14 @@ <h2>Version Inventory and Inventory Digest</h2>
             SHOULD include an inventory file that is an <a href="#inventory">Inventory</a> of all content for versions
             up to and including that particular version. Where an OCFL Object contains <code>inventory.json</code>
             in version directories, the inventory file in the OCFL Object Root MUST be the same as the file in the
-            most recent version. Every inventory file MUST have a corresponding <a href="#inventory-digest">Inventory
-            Digest</a>.
+            most recent version, and that inventory file MUST NOT be subsequently deleted. Every inventory file MUST
```
rosy1280: s/most recent version/most recent version directory

julianmorley:
Withdrawing in favor of more explicit verbiage elsewhere.

PR #394, opened 2019-10-04
Add Julian's orcid to spec and notes.

julianmorley:


ahankinson:
Sorry @awoods @rosy1280 -- i assigned reviewers and wiped out your reviews. (It's nice seeing the list of reviewers up top so we can see who has reviewed and who hasn't.)

PR #400, opened 2019-10-23
Expunge deposit directory references in spec

### LINKED ISSUE #396: Deposit directory
Am I right in remembering that this paragraph:

> An OCFL Storage Root may contain a directory named deposit, which may be empty. Implementations may use this directory for assembling new or updated content. Clients performing any other operation, or validating a storage root, must ignore the deposit directory if present.

should now be removed from the spec?

neilsjefferies:
Fixes #396

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -892,13 +892,8 @@ DIGEST inventory.json
           </li>
         </ul>
         <p>
-            An OCFL Storage Root MAY contain a directory named <code>deposit</code>, which MAY be empty. Implementations
-            MAY use this directory for assembling new or updated content. Clients performing any other operation, or
-            validating a storage root, MUST ignore the <code>deposit</code> directory if present.
-        </p>
-        <p>
-          Except for a <code>deposit</code> directory, sub-directories within an OCFL Storage Root MUST NOT contain
-          files that are not part of an OCFL Object. Empty directories MUST NOT appear within a storage root.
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -892,13 +892,8 @@ <h2>Root Structure</h2>
...
         <p>
-            An OCFL Storage Root MAY contain a directory named <code>deposit</code>, which MAY be empty. Implementations
-            MAY use this directory for assembling new or updated content. Clients performing any other operation, or
-            validating a storage root, MUST ignore the <code>deposit</code> directory if present.
-        </p>
-        <p>
-          Except for a <code>deposit</code> directory, sub-directories within an OCFL Storage Root MUST NOT contain
-          files that are not part of an OCFL Object. Empty directories MUST NOT appear within a storage root.
+          Sub-directories within an OCFL Storage Root MUST NOT contain files that are not part of an OCFL Object.
+          Empty directories MUST NOT appear within a storage root.
```
julianmorley: I think it's possible for a well-formed object to have an empty logs directory. Do we really mean empty directories must not appear within a version directory?

ahankinson: I think we can have empty logs directories. We can't have empty directories within a version because we can't track that in an inventory. 

We also can't have empty directories within a storage root that are not part of an OCFL Object.

But that means that there is really no problem with having empty directories that are within an OCFL Object but not within the version directories (logs or extensions).

zimeon:
See, I like it so much I double approved it!

Issue #370, opened 2019-09-16
Unversioned objects and mutable versions (split out from #367)


neilsjefferies:
It occurs to me that these two cases can be addressed by the same mechanism (to be decided). 

We allow an OCFL object to have a mutable head version designated by some means. Unlike something in the <code>deposit</code> directory, it must always be a valid OCFL version but can be changed by overwriting without creating a new version.

A unversioned object simply has *ony* a single mutable version. 

Implementation Notes should indicate that mutable versions should not be referenced explicitly by external systems (citation etc.) since they are not consistent. Instead such objects should be referenced at the ID level with no explicit version, or by reference to the latest stably defined vsersion.

Such versions would need more detailed Implementation Notes to cover safe updating.

Unversioned objects and objects with mutable versions should be considered a bit more "fragile" since state with repect to replication and referencing is less well defined. I think explicitly versioned persistence should remain the "gold" standard for good preservation.     

Purge redirects might also be considered as unversioned objects - it might make sense if multiple purge operations have happened.

zimeon:
I agree that is we allow mutability, a special mutable version separate from the regular versions (`v1`, `v2`, etc.) is important so that the mutability option doesn't drag the stronger OCFL versioning guarantees down.

Not sure we need to mention or define reference mechanisms. Except for saying that objects have ids we don't do this now.

zimeon:
https://ocfl.github.io/extensions/0005-mutable-head.html

PR #402, opened 2019-10-24
WIP DNM: Sample Notes for var implementation

### LINKED ISSUE #373: Clarify whether and OCFL Object Root may contain other files/directories not specified

neilsjefferies:
Assuming var is acceptable, this is a quick example of some relevant IN's. Fixes #373 and possibly #370

### DIFF
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -782,6 +782,93 @@
       </section>
     </section>
   </section>
+  <section>
+    <h2>Special Case Object</h2>
+    <p>
+    This section contains details of a number of object types that are useful to explore   
+    in more detail than the strict OCFL specification. All these objects are OCFL compliant
+    but can usefully be defined in a common manner to simplify application design and 
+    object portability.
+    </p>
+    <section>
+      <h2>Placeholder Objects</h2>
+      <p>
+      A placeholder object has no content but serves to register the allocation of an object
+      ID, and the corresponding location in the OCFL object root, to ensure that it cannot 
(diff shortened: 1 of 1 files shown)


zimeon:
In additional to making a decision on the underlying issue of whether we want to include the possibility of a `var` directory (#373), I think we need to have a discussion of how we talk about content in a `var` directory. My sense is that it may be dangerous to talk about it as a version, even and unmerged one. One might be able to use language around unmerged or uncommitted content that is can be described using a similar inventory structure to facilitate creation of a new version... or some such. But in OCFL terms it isn't an [OCFL Version](https://ocfl.io/draft/spec/#dfn-ocfl-version).

awoods:
@zimeon : Do you believe @neilsjefferies' language in this PR is tied too closely with "OCFL Versions"?

zimeon:
@awoods: yes, e.g. _"Mutable head objects have a current version"_

neilsjefferies:
Not needed now we have extensions mechanism

PR #404, opened 2019-10-29
WIP: Extensions

ahankinson:
A "commit early, commit often" take on the wording for the extensions directory. Still todo:

- [ ]  Add general 'extensions' description (top-level section?) to which we can refer as a general description of extensions and their mechanics.
- [ ]  Review requirements for extensions in Object root
- [ ]  Add language about storing plain text descriptions of the extension in the object root

Add any missing things you see to the list.

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -854,7 +854,17 @@ DIGEST inventory.json
                 wrong, while others may wish to only store a log entry if an intervention was required.
             </p>
         </blockquote>
+    </section>
 
+    <section id="extensions">
+        <h2>Extensions directory</h2>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
+            of extending the functionality of an OCFL Object. Sub-directories of the extensions directory MUST be named
+            according to a registered extension name. OCFL Validators MUST ignore any content in extensions that are
+            not understood. The specific structure and function of the extension, as well as a declaration of the
+            registered extension name MUST be defined in the OCFL Extensions Registry.
+        </p>
     </section>
 </section>
 
@@ -953,7 +963,7 @@ DIGEST inventory.json
     </section>
 
     <section id="root-extensions">
-        <h2>Extensions</h2>
+        <h2>Root Extensions</h2>
         <p>
             The behavior of the storage root may be extended to support features from other specifications.
             An OCFL validator MUST ignore any files in the storage root it does not understand. Additional
@@ -1512,6 +1522,27 @@ myfirstbag
 }
 </pre>
 </section>
+    <section id="example-extended-object">
+        <h2>Example Extended OCFL Object</h2>
+        <p>
+            The following example OCFL Object has an extension containing custom content. The OCFL Object itself
+            remains valid.
+        </p>
+<pre>
+[object root]
+    ├── 0=ocfl_object_1.0
+    ├── inventory.json
+    ├── inventory.json.sha512
+    ├── extensions
+    │   └── my-extension
+    │       └── file1-new.txt
+    └── v1
+        ├── inventory.json
+        ├── inventory.json.sha512
+        └── content
+            └── file.txt
+</pre>
+    </section>
 </section>
 </body>
 </html>


draft/spec/index.html
```diff
@@ -854,7 +854,17 @@ <h2>Logs Directory</h2>
...
         </blockquote>
+    </section>
 
+    <section id="extensions">
+        <h2>Extensions directory</h2>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
+            of extending the functionality of an OCFL Object. Sub-directories of the extensions directory MUST be named
+            according to a registered extension name. OCFL Validators MUST ignore any content in extensions that are
+            not understood. The specific structure and function of the extension, as well as a declaration of the
```
rosy1280: > not understood

that's a little unclear.  we should be a bit more specific.  i'm assuming you mean by a validator that understands the spec and/or that can validate whatever the extension asks it to validate.

draft/spec/index.html
```diff
@@ -953,7 +963,7 @@ <h2>Storage Hierarchies</h2>
     </section>
 
     <section id="root-extensions">
-        <h2>Extensions</h2>
+        <h2>Root Extensions</h2>
```
rosy1280: This makes it sound like their is a difference between extensions that appear in the storage root and the object root.  I thought we had agreed that Extensions would be stored at the root and then used in the objects -- but you might not use all extensions listed in the storage root.

ahankinson: There are extensions, like the “layout” functionality, that will be root-only. In this case it does not affect the contents of objects, but only the paths to them.

draft/spec/index.html
ahankinson: I took the language from the existing §4.4, "An OCFL validator must ignore any files in the storage root it does not understand."

We also have (in §3.4): "Optional fixity algorithms that are not supported by a client must be ignored by that client."

So would something like "Extension directories that are not supported by a client MUST be ignored by that client" make sense? Should we narrow it down specifically to validation, or just leave it open generally to "clients"?

zimeon: I think it should be "clients" and agree with last wording suggestion

draft/spec/index.html
```diff
@@ -854,7 +854,17 @@ <h2>Logs Directory</h2>
                 wrong, while others may wish to only store a log entry if an intervention was required.
             </p>
         </blockquote>
+    </section>
 
+    <section id="extensions">
+        <h2>Extensions directory</h2>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
```
zimeon: I would favor a shorter name, `exts` would be fine for example. I think this not only helps control path lengths but actually makes it easier to talk about the directory in specs vs the extensions that go in them

draft/spec/index.html
zimeon: Change title to "Storage Root Extensions"

draft/spec/index.html
```diff
@@ -854,7 +854,17 @@ <h2>Logs Directory</h2>
                 wrong, while others may wish to only store a log entry if an intervention was required.
             </p>
         </blockquote>
+    </section>
 
+    <section id="extensions">
+        <h2>Extensions directory</h2>
```
zimeon: Capital D in directory to match section title style

draft/spec/index.html
```diff
@@ -854,7 +854,17 @@ <h2>Logs Directory</h2>
...
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
+            of extending the functionality of an OCFL Object. Sub-directories of the extensions directory MUST be named
```
zimeon: * Here we do not make it illegal to have files in the extension directory, I think that is the intention
 * I would also say the sub-directories SHOULD be named according to a registered name rather then MUST

draft/spec/index.html
```diff
@@ -854,7 +854,17 @@ <h2>Logs Directory</h2>
...
+            according to a registered extension name. OCFL Validators MUST ignore any content in extensions that are
+            not understood. The specific structure and function of the extension, as well as a declaration of the
+            registered extension name MUST be defined in the OCFL Extensions Registry.
```
zimeon: I'd go for SHOULD instead of MUST again, I'd also put any such language next to the sentence about the name

draft/spec/index.html
zimeon: ... which still leave the question of where descriptions of object extensions go

draft/spec/index.html
ahankinson: My concern with `exts` is that it is possible to extrapolate different words ("extras"? "externals"?) and it's unlikely that six extra characters will make a difference in path length limits (if you're running that close to the edge you'll likely have other, larger problems). 

The unambiguous nature of the full word `extensions` seems to me to be clearer. We generally apply `code formatting` to talk about directory names to help differentiate between that and extensions in general.

draft/spec/index.html
ahankinson: Descriptions of object extensions also go in the storage root as a plain text file, on the theory that it is a single place that can describe the structure of some or all objects in that storage root.

draft/spec/index.html
ahankinson: SHOULD would leave open the possibility of registered and unregistered namespace collisions -- two extensions with the same name but different functions. By making it a MUST we reduce this possibility.

draft/spec/index.html
ahankinson: SHOULD would leave open the possibility of multiple things implementing the same functionality but with different local names. This would make it difficult for clients to understand that it can operate on that content. (The alternative would be to use a NamAsTe file to declare an extension's identity in the root directory of that extension, e.g., (`extensions/versioning/0=mutable_head-1.0`), which would allow us to name things differently. But this was an unpopular suggestion on the editorial call).

draft/spec/index.html
neilsjefferies: I think stick with validators - this is the spec for objects at rest. We can discuss clients in the Implementation Notes (Client Behaviour's section) - ignoring extentions when writing is dangerous.

draft/spec/index.html
neilsjefferies: Agree with @ahankinson

draft/spec/index.html
neilsjefferies: "ext" is a common enough root in enough languages that indicates outside-ness that I don't see much confusion. That being said, we have a mechanism for mapping long path names in the inventory anyway so it's not that big an issue.

draft/spec/index.html
neilsjefferies: Aren't the directories under <code>extensions</code> a namespace that references OCFL extentions - in that case, there should only be a single instance of any one type.

ahankinson: My concern is that by making it a SHOULD we leave open the possibility that people name their directories in `extensions` in a way that diverges from the name that is registered by the extension.

neilsjefferies: @ahankinson I'm agreeing with you!

awoods:
Superseded by: https://github.com/OCFL/spec/pull/428

PR #409, opened 2019-11-22
Explain fixity optional, for legacy, MAY be any subset of files

### LINKED ISSUE #408: Make it clear that the `fixity` block may not contain fixity for all files in an object
Split from #406: we need to make it clear that the `fixity` block may not contain fixity for all files in a version or object. The `manifest` serves this purpose, MUST be complete, and is the canonical fixity information for all content files. The `fixity` block is an optional convenience for storing _additional_ fixity values, typically from migrations. It is also optional whether `fixity` information is carried forward from one version to another.

zimeon:
Fixes #408

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -758,25 +758,28 @@
         <section id="fixity">
             <h2>Fixity</h2>
             <p>
-                An OCFL Object inventory MAY include a block for storing fixity checks. This block MUST have the
-                key of <code>fixity</code> within the inventory.
+                An OCFL Object inventory MAY include a block for storing additional fixity information to supplement the
+                complete set of digests in the <a href="#manifest">Manifest</a>, for example to support legacy digests
+                from a content migration. This block MUST have the key of <code>fixity</code> within the inventory.
             </p>
             <p>
                 The <code>fixity</code> block MUST contain keys corresponding to the controlled vocabulary given in
-                the <a href="#digest-algorithms">digest algorithms</a> listed in the Digests section, or in a table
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -758,25 +758,28 @@ <h2>Version</h2>
...
-                the <a href="#digest-algorithms">digest algorithms</a> listed in the Digests section, or in a table
-                given in an <a>Extension</a>. The value of the fixity block MUST follow the structure of the
-                <a href="#manifest"><code>manifest</code></a> block; that is, a key corresponding to
-                the digest value, and an array of <a>content path</a>s that match that digest.
+                the <a href="#digest-algorithms">digest algorithms</a> listed in the <a href="#digests">Digests</a>
+                section, or in a table given in an <a>Extension</a>. The value of the fixity block for a particular
+                digest algorithm MUST follow the structure of the <a href="#manifest"><code>manifest</code></a> block;
+                that is, a key corresponding to the digest value, and an array of <a>content path</a>s that match that
+                digest. The fixity block for any digest algorithm MAY include digest values for any subset of
+                content paths in the object.
```
ahankinson: Possibly we should add something like "There is no requirement that all files have a value in the fixity block" or something along those lines, just to make it clear that the fixity block may contain an incomplete set of hashes for the whole object.

zimeon: That was the intention of "The fixity block for any digest algorithm MAY include digest values for any subset of content paths in the object." but if we want to really drive that home we could  add the extra, or else repeat in implementation notes

zimeon: Thinking about this more, I agree that we should spell it out. Will add to the PR...

PR #423, opened 2020-02-18
Fixed: Logical paths clarification

### LINKED ISSUE #411: Logical Paths: Allowed Characters and Path Separators
To address #407 the editors have decided to restrict characters in the **content paths**, but have not reached agreement on what can be in the **logical paths**. The options are:

 - Maintain the same restrictions as the content paths (aka no Windows-sensitive names)
 - Have no restrictions on the names INCLUDING that both forward and back slashes are allowed as logical names, depending on the system that was used to ingest the content.
 - Have no restrictions on the names EXCEPT that path separators must be normalized to forward (/) slashes.

The editors were in agreement that option 1 is not necessary. There was disagreement on points two and three, so this issue is to provide a space for additional feedback on the issue.

### LINKED ISSUE #413: Add section 3.3.2 file path names
add section 3.3.2 to the spec.  
- [ ] discuss allowable characters in file names for the content path
- [ ] clearly state that file name restrictions identified in this section don't apply to the logical path.

for something similar take a look at [section 6.1.1.3 of the BagIt spec](https://tools.ietf.org/html/rfc8493#section-6.1.1.3)

ahankinson:
Clarify that logical paths must not start with reserved characters. Also adds 'and path' in the definition of logical paths.

Fixes #411 
Fixes #413

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -296,7 +296,8 @@
         </dt>
         <dd>
             A path that represents a file's location in the <a>logical state</a> of an object. Logical paths are
-            used in conjunction with a digest to represent the file name for a given bitstream at a given version.
+            used in conjunction with a digest to represent the file name and path for a given bitstream at a given
+            version.
         </dd>
         <dt>
             <dfn>Logical State</dfn>:
@@ -696,7 +697,8 @@
                             are digest values, each of which MUST correspond to an entry in the
                             <a href="#manifest">manifest of the inventory</a>. The value for each key is an array
                             containing <a>logical path</a> names of files in the OCFL Object state that have
-                            content with the given digest.
+                            content with the given digest. Logical paths MUST NOT start with <code>.</code>,
+                            <code>..</code>, <code>/</code>, or <code>//</code>.
                         </p>
                         <blockquote class="informative">
                             <p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -702,7 +702,7 @@
                         <p>
                             <a href="#logical-path">Logical paths</a> present the structure of an OCFL Object at a
                             given version. This is given as an array of file paths, with the following restrictions
-                            provide for path safety in this common case. The logical path SHOULD be interpreted as a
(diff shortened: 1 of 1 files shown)


ahankinson:
I agree that the SHOULD become MUST, but before doing that I would want to hear from @julianmorley

rosy1280:
this also looks like it will close #413

draft/spec/index.html
```diff
@@ -698,6 +699,14 @@ <h2>Version</h2>
                             containing <a>logical path</a> names of files in the OCFL Object state that have
                             content with the given digest.
                         </p>
+                        <p>
+                            <a href="#logical-path">Logical paths</a> present the structure of an OCFL Object at a
+                            given version. This is given as an array of file paths, with the following restrictions
```
julianmorley: consider "This is given as an array of values, with the following restrictions to provide for path safety in the common case of the logical path value representing a file path."

zimeon: Nice wording. I think we have to be strict to cover that case but I agree that it may not be used in that way

draft/spec/index.html
```diff
@@ -698,6 +699,14 @@ <h2>Version</h2>
...
+                            provide for path safety in this common case. The logical path MUST be interpreted as a
```
julianmorley: If I was being particularly bolshy, I'd say "The logical path MUST be interpreted as a set of one or more elements joined by a <code>/</code> separator. Elements MUST NOT be <code>.</code>, <code>..</code>, or empty (<code>//</code>)."

draft/spec/index.html
zimeon: At that point aren't we just obfuscating?

draft/spec/index.html
zimeon: At that point aren't we just obfuscating?

ahankinson: I think the current wording is a bit clearer, even if it uses more words.

draft/spec/index.html
ahankinson: I have incorporated Julian's suggested revision into the latest version

PR #425, opened 2020-02-18
Describe inventory consistency between versions

### LINKED ISSUE #421: Should we say anything about consistency or otherwise of version metadata between version inventories?
Imagine an object with `v1` and `v2`. In each of `v1/inventory.json` and `v2/inventory.json` there is a block `"versions": { "v1": { "created": ..., "message": ..., "state": ...., "user": ... } }`.

I feel that it is implied that the `state` must give a consistent set of files for `v1` in each case (though the values may be different if different digest algorithms are used for different versions). However, do the values of `created`, `message` and `user` need to (MUST) be consistent?  Or is changing them an allowed but discouraged (SHOULD) way to fix metadata in a system that might have immutable versions? There is currently no comment in https://ocfl.io/draft/spec/#version-inventory

zimeon:
Fixes #421

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -834,6 +834,11 @@ DIGEST inventory.json
             most recent version. Every inventory file MUST have a corresponding <a href="#inventory-digest">Inventory
             Digest</a>.
         </p>
+        <p>
+            In the case that prior version directories include an inventory file there will be multiple inventory 
+            files describing prior versions within the OCFL Object. Each version block in each prior inventory file MUST represent the same object state as the corresponding version block in the current inventory file. 
+            Additionally, the values of the <code>created</code>, <code>message</code> and <code>user</code> keys in each version block in each prior inventory file SHOULD have the same values as the corresponding keys in the correspodings version block in the current inventory file.
+        </p>
         <blockquote class="informative">
             Non-normative note: Storing an inventory for every version provides redundancy for this critical
             information in a way that is compatible with storage strategies that have immutable version directories.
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -834,6 +834,14 @@ <h2>Version Inventory and Inventory Digest</h2>
             most recent version. Every inventory file MUST have a corresponding <a href="#inventory-digest">Inventory
             Digest</a>.
         </p>
+        <p>
+            In the case that prior version directories include an inventory file there will be multiple inventory 
+            files describing prior versions within the OCFL Object. Each version block in each prior inventory file
+            MUST represent the same object state as the corresponding version block in the current inventory file. 
+            Additionally, the values of the <code>created</code>, <code>message</code> and <code>user</code> keys in
+            each version block in each prior inventory file SHOULD have the same values as the corresponding keys in
+            the correspodings version block in the current inventory file.
```
ahankinson: Is this saying that it is possible to change the responsibility of previous versions?

ahankinson: typo in 'corresponding' L843

zimeon: Thanks, typo fixed.

This doesn't talk about change, just the state of different version inventories at any point in time. Having said that, I think of it as a rule going forward in time -- I SHOULD NOT change the v1 message in the v2 inventory

ahankinson: I guess I'm asking why we're saying we "SHOULD NOT" change the message in v1 when it's referenced in the v2 inventory, instead of "MUST NOT". I would think that all versions of the v1 block in all inventories MUST be exactly the same.

zimeon: Well, that goes back to the question of #421 -- I prefer SHOULD because I can see legitimate reasons for an update yet one would want to ever rewrite the old version inventory

ahankinson: What legitimate reasons are there for updating the `created`, `message`, and `user` values of previous versions?

zimeon: Will move discussion back up to #421...

Issue #403, opened 2019-10-29
Add 'extensions' support to spec


awoods:
Per the Editors Meeting on 2019-10-29, this task is to add language to the OCFL specification that details support for optional extensions.

The language should include:
- description of what goes in the Storage Root (including naming conventions)
- description of what goes in the Object Root (including naming conventions)
- description of scope and limitation of extensions (i.e. can not change other elements of an OCFL Object)

We may be able to repurpose some of the language currently found in the `log` description.

rosy1280:
related to WIP PR: #404

awoods:
Additional reference: https://github.com/OCFL/spec/wiki/2020.01.21-Editors-Meeting

awoods:
Before proposing specification language, I would like to establish consensus on the details of what we are attempting to describe and define as it relates to "extensions".

1. Types of extensions
   1. Extensions referenced in specification, examples:
      - ocfl_layout.json
      - digest algorithms
   1. Extensions that require a workspace, under a subdirectory of 'extensions/', example:
      - mutable head
1. Questions to be answered
   1. How to reference extensions?
      - As sub-directory of 'extensions' directory, sharing name of extension definition
      - As a key in a specification-defined file (fixity-block, ocfl_layout.json)
   1. Where is extension referenced?
      - In 'ocfl_layout.json'
      - In 'fixity-block'
      - As sub-directory of 'extensions' directory
   1. Where does extension definition live?
      - In extensions repository
      - In the Storage Root
      - Naming convention: 4-digit, zero-padded prefix; should use hyphens to separate words; and have the .md extension.
   1. Where is extension workspace?
      - In sub-directory of 'extensions' directory of Storage Root and/or Object Root
      - Sub-directory must share the extension name, minus the '.md' suffix

zimeon:
IMO in point one we don't want to go into much detail of what type of extension there may be as that would appear to bless some directions and speak against others.  I thought this ticket was specifically for description of the `extensions` directory and this not really about digest algorithms and such which are already described. In order to get something that is readily mergeable I would be tempted to rename 4.4 to Storage Root Extensions and add a new 3.X Object Extensions, then later think about what common language or cross-reference there should be

PR #428, opened 2020-02-25
Add text defining storage root and object root 'extensions' directories

awoods:
Resolves: https://github.com/OCFL/spec/issues/403

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -868,21 +868,21 @@ DIGEST inventory.json
         <p>
             The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
             of extending the functionality of an OCFL Object. The <code>extensions</code> directory MUST NOT contain
-            any files, and no sub-directories other than extension sub-directories. Extension sub-directories MUST be
+            any files, and no sub-directories other than extension sub-directories. Extension sub-directories SHOULD be
             named according to a registered extension name. The specific structure and function of the extension, as
             well as a declaration of the registered extension name MUST be defined in one of the following locations:
-            <ul>
-              <li>
-                The <a href="https://ocfl.github.io/extensions/">OCFL Extensions repository</a>
-              </li>
-              <li>
-                The Storage Root as a plain text document
-              </li>
-            </ul>
         </p>
+        <ul>
+          <li>
+            The <a href="https://ocfl.github.io/extensions/">OCFL Extensions repository</a>
+          </li>
+          <li>
+            The Storage Root, as a plain text document directly in the Storage Root
+          </li>
+        </ul>
         <blockquote class="informative">
           <p>
-            Non-normative note: Extension sub-directories are required to use the same name as the registered extension
+            Non-normative note: Extension sub-directories should use the same name as a registered extension
             in order to both avoid the possiblity of an extension sub-directory colliding with the name of another
             registered extension as well as to facilitate the recognition of extensions by OCFL consumer clients.
           </p>
@@ -985,9 +985,9 @@ DIGEST inventory.json
             The behavior of the storage root may be extended to support features from other specifications.
         </p>
         <p>
-            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
-            of extending the functionality of an OCFL Object. The storage root <code>extensions</code> directory MUST
-            conform to the same guidelines and limitations as those defined for
+            The base directory of an OCFL Storage Root MAY contain a directory named <code>extensions</code> for the
+            purposes of extending the functionality of an OCFL Storage Root. The storage root <code>extensions</code>
+            directory MUST conform to the same guidelines and limitations as those defined for
             <a href="#object-extensions">object extensions</a>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -954,10 +979,18 @@ <h2>Storage Hierarchies</h2>
...
 
-    <section id="root-extensions">
-        <h2>Extensions</h2>
+    <section id="storage-root-extensions">
+        <h2>Storage Root Extensions</h2>
         <p>
             The behavior of the storage root may be extended to support features from other specifications.
+        </p>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
```
zimeon: s/OCFL Object/OCFL Storage Root/

draft/spec/index.html
```diff
@@ -954,10 +979,18 @@ <h2>Storage Hierarchies</h2>
...
+            of extending the functionality of an OCFL Object. The storage root <code>extensions</code> directory MUST
```
zimeon: s/OCFL Object/OCFL Storage Root/

draft/spec/index.html
```diff
@@ -1514,6 +1547,27 @@ <h2>Moab in an OCFL Object</h2>
...
+            The following example OCFL Object has an extension containing custom content. The OCFL Object itself
+            remains valid.
+        </p>
+<pre>
+[object root]
+    ├── 0=ocfl_object_1.0
+    ├── inventory.json
+    ├── inventory.json.sha512
+    ├── extensions
+    │   └── 0011-my-extension
```
zimeon: Do we need to create an example extension, maybe `9999-example-extension` or such in order for this to be valid?

draft/spec/index.html
```diff
@@ -861,7 +861,32 @@ <h2>Logs Directory</h2>
...
 
+    <section id="object-extensions">
+        <h2>Object Extensions</h2>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
+            of extending the functionality of an OCFL Object. The <code>extensions</code> directory MUST NOT contain
+            any files, and no sub-directories other than extension sub-directories. Extension sub-directories MUST be
+            named according to a registered extension name. The specific structure and function of the extension, as
+            well as a declaration of the registered extension name MUST be defined in one of the following locations:
+            <ul>
```
zimeon: Need to close the `<p>` block before opening `<ul>`, this is the reason for the travis error (and, of course, remove `<\p>` after)

draft/spec/index.html
```diff
@@ -861,7 +861,32 @@ <h2>Logs Directory</h2>
...
             </p>
         </blockquote>
+    </section>
 
+    <section id="object-extensions">
+        <h2>Object Extensions</h2>
+        <p>
+            The base directory of an OCFL Object MAY contain a directory named <code>extensions</code> for the purposes
+            of extending the functionality of an OCFL Object. The <code>extensions</code> directory MUST NOT contain
+            any files, and no sub-directories other than extension sub-directories. Extension sub-directories MUST be
```
zimeon: Per https://github.com/OCFL/spec/wiki/2020.01.21-Editors-Meeting I think agreement was that:
  * "Extensions SHOULD be a defined/registered"
  * "Extensions MUST be in their own sub-directory"

zimeon:
A few minor issues above but otherwise I think this looks great!

draft/spec/index.html
rosy1280: we have an example one `0000-example-extension` see: https://github.com/OCFL/extensions/blob/master/docs/0000-example-extension.md

draft/spec/index.html
```diff
@@ -861,7 +861,32 @@ <h2>Logs Directory</h2>
...
+            named according to a registered extension name. The specific structure and function of the extension, as
+            well as a declaration of the registered extension name MUST be defined in one of the following locations:
+            <ul>
+              <li>
+                The <a href="https://ocfl.github.io/extensions/">OCFL Extensions repository</a>
+              </li>
+              <li>
+                The Storage Root as a plain text document
```
ahankinson: I can see this leading to confusion that object extensions can be put in the storage root `extensions` directory, not directly in the storage root itself.

draft/spec/index.html
zimeon: Thanks @rosy1280 , I'd forgotten that. @awoods change looks good so will resolve

draft/spec/index.html
```diff
@@ -861,7 +861,32 @@ <h2>Logs Directory</h2>
...
+          </li>
+          <li>
+            The Storage Root, as a plain text document directly in the Storage Root
+          </li>
+        </ul>
+        <blockquote class="informative">
+          <p>
+            Non-normative note: Extension sub-directories should use the same name as a registered extension
+            in order to both avoid the possiblity of an extension sub-directory colliding with the name of another
+            registered extension as well as to facilitate the recognition of extensions by OCFL consumer clients.
```
ahankinson: remove 'consumer'? We don't really define what that is.

draft/spec/index.html
```diff
@@ -1514,6 +1547,27 @@ <h2>Moab in an OCFL Object</h2>
 }
 </pre>
 </section>
+    <section id="example-extended-object">
```
ahankinson: Perhaps we need an example Storage Root layout to clarify the extensions mechanism there?

draft/spec/index.html
zimeon: agree to drop "consumer"

draft/spec/index.html
zimeon: :+1:

PR #431, opened 2020-03-17
Change from ISO8601 to RFC3339

### LINKED ISSUE #419: Date format -- specification out of date
Our currrent reference to ISO8601 is https://ocfl.io/draft/spec/#bib-ISO8601 --> http://www.iso.org/iso/catalogue_detail?csnumber=40874 which is withdrawn by ISO and replaced by https://www.iso.org/standard/70907.html (2019 version).

However, I really don't like using the ISO spec because it isn't openly available. It is also likely more complex than we need. I wonder whether we'd be better with with [W3C Datetime](https://www.w3.org/TR/NOTE-datetime) (although that isn't really a spec) or [RFC3339](https://tools.ietf.org/html/rfc3339)?

zimeon:
Fixes #419

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -684,9 +684,10 @@
                         <code>created</code>
                     </dt>
                     <dd>
-                        The value of this key is the datetime of creation of this version. It MUST be expressed in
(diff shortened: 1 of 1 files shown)


zimeon:
Note that last two lines of change are just removing trailing spaces (which my editor did on its own)

PR #441, opened 2020-04-03
Add span tags with ids for RFC2119 terms

### LINKED ISSUE #440: Add error and warning ids to RFC2119 terms
From discussion on slack, add ids E001, E002, etc. to MUST and MUST NOT in the spec, and ids W001, W002 to SHOULD and SHOULD terms.

zimeon:
Fixes #440

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -410,78 +410,85 @@
                └── ... content files ...
 </pre>
         <p>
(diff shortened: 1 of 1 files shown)


zimeon:
Just for the record an in case I need to re-apply: code to add  https://gist.github.com/zimeon/670d6f55e53acd986b89accc69a42fe0

PR #442, opened 2020-04-07
Add first 15 errors and the 15 warnings according to new spec links

zimeon:
A start at rewriting the validation codes using the new spec links and text largely extracted from the spec (with context added in some cases)


zimeon:
Rendered page: https://github.com/OCFL/spec/blob/first-15-errors-warnings/validation/validation-codes.md

zimeon:
Closing in favor of #445 -- I'll add a new PR for the warnings

PR #444, opened 2020-04-07
Add E092.

julianmorley:
Tiny tweak to add a validation error that was missed in the initial pass.

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -831,10 +831,10 @@
         <h2>Inventory Digest</h2>
         <p>
             Every occurrence of an inventory file <span id="E058">MUST</span> have an accompanying sidecar file stating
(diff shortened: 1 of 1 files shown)


zimeon:
Well spotted!  I wonder whether we would best best to restructure the first sentence to include the name of the file in the MUST for E058, otherwise that error will be basically uncheckable. We could say:

> Every occurrence of an inventory file `[E058]`MUST have an accompanying sidecar file named `inventory.json.ALGORITHM` stating its digest, where <code>ALGORITHM</code> is the chosen digest algorithm for the object. The `ALGORITHM` `[E059]`MUST match the value given for the `digestAlgorithm` key in the inventory. An example might be `inventory.json.sha512`.

PR #445, opened 2020-04-08
Object Error codes

julianmorley:
A first pass at all the object validation error codes in the current draft.


zimeon:
Rendered version: https://github.com/OCFL/spec/blob/RFC2119/validation/validation-codes.md

zimeon:
Part of #446

Issue #446, opened 2020-04-08
Update table of validator codes to use id tags for RFC2119 terms


zimeon:
Follow on from #440 we want the set of [validation codes](https://github.com/OCFL/spec/blob/master/validation/validation-codes.md) to match

zimeon:
Still need to add error codes #069 through #091

awoods:
Resolved by: https://github.com/OCFL/spec/pull/471

PR #447, opened 2020-04-08
Update warning codes and descriptions

zimeon:
Follow style of #445 but just do the warnings. Part of #446


zimeon:
Rendered version is https://github.com/OCFL/spec/blob/rfc2119-warnings/validation/validation-codes.md#warnings-corresponding-with-should-in-specification

PR #452, opened 2020-04-21
Rephrase to remove redundant E092

### LINKED ISSUE #448: Rewrite description of sidecar file to remove uncheckable MUST
Per #444 we end up with 3 RFC2119 MUST for [describing inventory sidecar files](https://ocfl.io/draft/spec/#inventory-digest). I think we need only 2 MUST and propose changing the text to something like:

> Every occurrence of an inventory file [E058]MUST have an accompanying sidecar file named inventory.json.ALGORITHM stating its digest, where ALGORITHM is the chosen digest algorithm for the object. The ALGORITHM [E059]MUST match the value given for the digestAlgorithm key in the inventory. An example might be inventory.json.sha512.

which has the effect of removing E092

zimeon:
Fixes #448

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -830,10 +830,9 @@
     <section id="inventory-digest">
         <h2>Inventory Digest</h2>
         <p>
-            Every occurrence of an inventory file <span id="E058">MUST</span> have an accompanying sidecar file stating
-            its digest. This sidecar file <span id="E092">MUST</span> be of the form
-            <code>inventory.json.ALGORITHM</code>, where <code>ALGORITHM</code> is the chosen digest algorithm
-            for the object. This value <span id="E059">MUST</span> match the value given for the
+            Every occurrence of an inventory file <span id="E058">MUST</span> have an accompanying sidecar file named
+            <code>inventory.json.ALGORITHM</code> stating its digest, where <code>ALGORITHM</code> is the chosen digest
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -830,10 +830,9 @@ <h2>Fixity</h2>
     <section id="inventory-digest">
         <h2>Inventory Digest</h2>
         <p>
-            Every occurrence of an inventory file <span id="E058">MUST</span> have an accompanying sidecar file stating
-            its digest. This sidecar file <span id="E092">MUST</span> be of the form
```
awoods, line 834: Does the corresponding validation code also go away? 
   - E092 - https://github.com/OCFL/spec/blob/master/validation/validation-codes.md

zimeon: Yes _in theory_, but you'll note that E092 isn't in the set of validation codes yet, they haven't been completed that far

awoods: Hmm... I may be looking at the wrong file, but searching for "E092" in https://github.com/OCFL/spec/blob/master/validation/validation-codes.md shows an entry between "E058" and "E059".

zimeon: It does, how odd! I looked for the last entry which was just E068... have created https://github.com/OCFL/spec/pull/454

PR #457, opened 2020-05-05
Rework E039, remove duplicate W006

### LINKED ISSUE #450: E039 & W006 are duplicates of E025 & W004 - change language to avoid?
For content addressing, as expressed by `digestAlgorithm` we require `sha512` or `sha256` and say it SHOULD be `sha512`. This is described in https://ocfl.io/draft/spec/#E025 and https://ocfl.io/draft/spec/#W004

> For content-addressing, OCFL Objects MUST use either `sha512` or `sha256`, and SHOULD use `sha512`.

and then duplicated in https://ocfl.io/draft/spec/#E039 and https://ocfl.io/draft/spec/#W006

> This MUST be either sha512 or sha256, and SHOULD be sha512.

We have another small disconnect in that `digestAlgorithm` is not described as giving the algorithm for content-addressing in the way we discuss in the Digests section. There is also a typo in:

> See **thead** section on Digests for more information.

zimeon:
Fixes #450

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -620,10 +620,10 @@
                 </dd>
                 <dt><code>digestAlgorithm</code></dt>
                 <dd>
-                    The digest algorithm used for calculating digests within the OCFL Object. This
-                    <span id="E039">MUST</span> be either <code>sha512</code> or <code>sha256</code>, and
-                    <span id="W006">SHOULD</span> be <code>sha512</code>. See thead
-                    <a href="#digest-algorithms">section on Digests</a> for more information.
+                    The digest algorithm used for calculating digests for content-addressing within the OCFL Object
+                    and for the <a href="#inventory-digest">Inventory Digest</a>. This <span id="E039">MUST</span> be
+                    the algorithm used in the <code>manifest</code> and <code>state</code> blocks, see the
+                    <a href="#digests">section on Digests</a> for more information about algorithms.
                 </dd>
                 <dt><code>head</code></dt>
                 <dd>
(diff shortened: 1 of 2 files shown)


draft/spec/index.html
```diff
@@ -620,10 +620,10 @@ <h2>Basic Structure</h2>
                 </dd>
                 <dt><code>digestAlgorithm</code></dt>
                 <dd>
-                    The digest algorithm used for calculating digests within the OCFL Object. This
-                    <span id="E039">MUST</span> be either <code>sha512</code> or <code>sha256</code>, and
-                    <span id="W006">SHOULD</span> be <code>sha512</code>. See thead
-                    <a href="#digest-algorithms">section on Digests</a> for more information.
+                    The digest algorithm used for calculating digests for content-addressing within the OCFL Object
+                    and for the <a href="#inventory-digest">Inventory Digest</a>. This <span id="E039">MUST</span> be
+                    the algorithm used in the <code>manifest</code> and <code>state</code> blocks, see the
```
awoods, line 625: The digest algorithm is used for:
- content addressing,
- inventory digests, *and*
- the fixity block

Did you intentionally omit the fixity block usage?

zimeon: Yes, because the fixity uses an independently defined and likely different set of digest algorithms. The fixity block use is not content addressing

draft/spec/index.html
```diff
@@ -620,10 +620,10 @@ <h2>Basic Structure</h2>
                 </dd>
                 <dt><code>digestAlgorithm</code></dt>
                 <dd>
-                    The digest algorithm used for calculating digests within the OCFL Object. This
-                    <span id="E039">MUST</span> be either <code>sha512</code> or <code>sha256</code>, and
-                    <span id="W006">SHOULD</span> be <code>sha512</code>. See thead
-                    <a href="#digest-algorithms">section on Digests</a> for more information.
+                    The digest algorithm used for calculating digests for content-addressing within the OCFL Object
```
ahankinson, line 623: Is content addressing really hyphenated? Seems strange

zimeon: It is 3 out of the 4 other times it appears in the spec so I'm being consistent

awoods: It is probably one of those things that could go either way. There is something to be said for consistency... and it appears that wikipedia goes with the hyphen:
https://en.wikipedia.org/wiki/Content-addressable_storage

Issue #461, opened 2020-05-05
Clarify issues with case insensitive digests resulting in repeated digests


zimeon:
Consider https://github.com/OCFL/fixtures/tree/master/1.0/bad-objects/bad16_digest_repeated which has inventory:

```
{
  "digestAlgorithm": "sha512",
  "head": "v1",
  "id": "ark:123/abc",
  "manifest": {
    "cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e": [
      "v1/content/empty1"
    ],
    "CF83E1357EEFB8BDF1542850D66D8007D620E4050B5715DC83F4A921D36CE9CE47D0D13C5D85F2B0FF8318D2877EEC2F63B931BD47417A81A538327AF927DA3E": [
      "v1/content/empty2"
    ]
  },
  "type": "https://ocfl.io/1.0/spec/#inventory",
  "versions": {
    "v1": {
      "created": "2019-01-01T02:03:04Z",
      "message": "Example with two empty files that thus have the same digest but are represented with different normalizations",
      "state": {
        "cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e": [
          "empty1"
        ],
        "CF83E1357EEFB8BDF1542850D66D8007D620E4050B5715DC83F4A921D36CE9CE47D0D13C5D85F2B0FF8318D2877EEC2F63B931BD47417A81A538327AF927DA3E": [
          "empty2"
        ]
      }
    }
  }
}
```

the digests are actually the same but in both `manifest` and `state` one is in uppercase and the other in lowercase. This could be allowed (should be text explaining merge after normalization) or it could be forbidden (what I'd assumed, should be text explaining that different normalizations MUST not result in repeated digests)

zimeon:
See #437 where we decided to keep case insensitivity

awoods:
Given the previous decision (https://github.com/OCFL/spec/issues/437#issuecomment-606625847), do we want to add "text explaining merge after normalization", or do nothing?

zimeon:
I'm inclined to require consistent normalization within a single `inventory.json` but I think we have to add explanation either way

neilsjefferies:
[...]

PR #468, opened 2020-05-15
Digests must be unique in manifest and fixity

neilsjefferies:
Addresses #461

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -654,7 +654,9 @@
                 of every content file in all versions of the <a>OCFL Object</a>. The value for each key
                 <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the OCFL
                 Object that have content with the given digest. Content paths within a manifest block
-                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>.
+                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>. As JSON keys are case 
+                sensitive, while digests may not be, there is an additional requirement that each digest value 
+                <span id="E096">MUST</span> occur only once in the manifest regardless of case.
             </p>
             <blockquote class="informative">
                 <p>
@@ -810,7 +812,10 @@
                 values for any subset of content paths in the object. Where included, the digest values given
                 <span id="E093">MUST</span> match the digests of the files at the corresponding content paths. There
                 is no requirement that all content files have a value in the fixity block, or that fixity values
-                provided in one version are carried forward to later versions.
+                provided in one version are carried forward to later versions. As JSON keys are case 
+                sensitive, while digests may not be, there is an additional requirement that each digest value 
+                <span id="E097">MUST</span> occur only once in the fixity block for any digest algorithm, regardless
+                of case.
             </p>
             <blockquote class="informative">
             <p>
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -653,10 +653,10 @@
                 The value of the <code>manifest</code> key is a JSON object, with keys corresponding to the digests
                 of every content file in all versions of the <a>OCFL Object</a>. The value for each key
                 <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the OCFL
-                Object that have content with the given digest. Content paths within a manifest block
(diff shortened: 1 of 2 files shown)


draft/spec/index.html
```diff
@@ -654,7 +654,9 @@ <h2>Manifest</h2>
                 of every content file in all versions of the <a>OCFL Object</a>. The value for each key
                 <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the OCFL
                 Object that have content with the given digest. Content paths within a manifest block
-                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>.
+                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>. As JSON keys are case 
+                sensitive, while digests may not be, there is an additional requirement that each digest value
```
ahankinson, line 658: "may not be" --> "are not" since we specify that digests are case insensitive.

neilsjefferies: Not all digests are ncessarily case insensitive and we don't actually require it as far as I can see. We just note that it is the case for the base16 digests that we currently specify.

zimeon: I think "may not be" is correct. Some digests might be case sensitive even if that doesn't apply to the ones we have specified so far. I would leave it as is.

zimeon:
This PR seems good as far as it goes but there are two more issues that it does not address:
* Should a digest also appear only once in a `state` block?
* I assume the sense is that the case in `manifest` and case in `state` do not have to match, I think we should note that

zimeon:
We could always merge this PR and have a new PR for the other parts, but I think they are still part of #461

draft/spec/index.html
neilsjefferies: ...but it IS awkward to validate if we can't assume that digests are case insensitive. Which would mean modifying the language around digests.

zimeon: IMO, every digest needs its own normalization scheme. That is what I built into my code

validation/validation-codes.md
```diff
@@ -82,6 +82,8 @@
 | E093 | 'Where included in the fixity block, the digest values given must match the digests of the files at the corresponding content paths.' | https://ocfl.io/draft/spec/#E093
 | E094 | 'The value of [the message] key is freeform text, used to record the rationale for creating this version. It must be a JSON string.' | https://ocfl.io/draft/spec/#E094
 | E095 | 'Within a version, logical paths must be unique and non-conflicting, so the logical path for a file cannot appear as the initial part of another logical path.' | https://ocfl.io/draft/spec/#E095
+| E096 | 'As JSON keys are case sensitive, while digests may not be, there is an additional requirement that each digest value must occur only once in the manifest regardless of case.' | https://ocfl.io/draft/spec/#E096
+| E097 | 'As JSON keys are case sensitive, while digests may not be, there is an additional requirement that each digest value must occur only once in the fixity block for any digiest algorithm, regardless of case.' | https://ocfl.io/draft/spec/#E097
```
ahankinson, line 86: typo "digest"

awoods:
@zimeon : Regarding:
> Should a digest also appear only once in a state block?

Given the statement in the `state` block description:
> The keys of this JSON object are digest values, each of which must correspond to an entry in the manifest of the inventory.

..maybe it is a non-issue? We could clarify what we mean by "correspond" to note that the case must match.

zimeon:
@awoods I think we need to be more specific that the keys in `state` must be exact string matches to the digest values in the `manifest`

validation/validation-codes.md
neilsjefferies: Good spot! Can't even cut-and-paste competently!

draft/spec/index.html
```diff
@@ -654,7 +654,9 @@ <h2>Manifest</h2>
                 of every content file in all versions of the <a>OCFL Object</a>. The value for each key
                 <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the OCFL
                 Object that have content with the given digest. Content paths within a manifest block
-                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>.
+                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>. As JSON keys are case
```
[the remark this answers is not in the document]
neilsjefferies: Agreed, and something similar for the fixity section too.

PR #472, opened 2020-05-19
Define rules for content paths

### LINKED ISSUE #462: Logical paths MUST not have element . .. or //, but what about content paths?
It seems we specify a number of restrictions on logical paths so that an object can be unpacked on a unix-like filesystem (https://ocfl.io/draft/spec/#version):

> The logical path must be interpreted as a set of one or more path elements joined by a / path separator. Path elements must not be ., .., or empty (//). Additionally, a logical path MUST NOT begin or end with a forward slash (/).

but we don't actually say anything comparable about content paths (https://ocfl.io/draft/spec/#manifest). I think currently the following is not disallowed:

```
"manifest": {
    "7dcc35...c31": [ "v1/content/foo/down/../bar.xml" ],
    "cf83e1...a3e": [ "v1/content/////empty.txt" ],
    "ffccf6...62e": [ "v1/content/./././././image.tiff" ]
  }
```

zimeon:
Closes #462

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -656,8 +656,20 @@
                 Object that have content with the given digest. As JSON keys are case sensitive, while digests may 
                 not be, there is an additional requirement that each digest value <span id="E096">MUST</span> occur
                 only once in the manifest regardless of case. Content paths within a manifest block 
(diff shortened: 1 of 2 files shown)


zimeon:
Thanks @neilsjefferies ... fixed

PR #476, opened 2020-05-28
Clarify rules for files and sub-directories under storage root

### LINKED ISSUE #470: Why does the Storage Root extensions section talk about what files are allowed in the root itself?
We have in https://ocfl.io/draft/spec/#storage-root-extensions

> An OCFL validator MUST ignore any files in the storage root it does not understand. Additional files MUST not appear in other directories under the storage root.

This isn't anything to do with extensions and:

1. if we want the first MUST is should be in https://ocfl.io/draft/spec/#root-structure
2. the second sentence is duplicative of https://ocfl.io/draft/spec/#E072 and should be removed

zimeon:
Suggestion to fix #470 (understanding that discussion may not yet be complete)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -967,9 +967,10 @@ DIGEST inventory.json
           <a href="#root-conformance-declaration">Root Conformance Declaration</a> identifying it as such.
         </p>
         <p>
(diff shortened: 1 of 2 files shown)


zimeon:
Thanks @neilsjefferies , fixed

PR #479, opened 2020-06-15
Inventory version must match object version

### LINKED ISSUE #475: Object and inventory versioning
The discussion in #474 raised a number of additional versioning related questions.

1. Is there a distinction between the object conformance declaration and the inventory `type` definition? That is, does one represent the version of the object and the the other the inventory, or do they both represent the version of the object as a whole?
1. MUST the OCFL version in the object conformance declaration match the version in the root inventory `type` definition?
1. Can an object created in one version of the spec later be updated to a different version?
1. If the OCFL version in an inventory and the object conformance declaration versions can change, independently or otherwise, then the versions in historic inventories may no longer match the object conformance declaration. Is this a problem? Should there be a namaste file in each version directory?

zimeon:
Fixes #475

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -616,8 +616,10 @@
                 <dt><code>type</code></dt>
                 <dd>
                     A type for the inventory JSON object that also serves to document the OCFL specification version
-                    that the inventory complies with. This <span id="E038">MUST</span> be the URI of the inventory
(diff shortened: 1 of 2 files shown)


zimeon:
@neilsjefferies the change of text from the spec to error is from "this" to "the type value", so don't we want all three words in [..]?

PR #481, opened 2020-06-16
Remove ocfl_layout.json

### LINKED ISSUE #477: Referencing extensions in ocfl_layout.json
When describing `ocfl_layout.json`, the draft OCFL spec reads:

> key - A key identifying the precise arrangement of directories and OCFL objects under the storage root, i.e. how OCFL object identifiers are mapped to directory hierarchies. The value of this key is not defined in the OCFL specification, but MUST correspond to a value given in an Extension.

The text is unclear what "value" the key must correspond to. Opening the issue here rather than under extensions because my assumption, which may be incorrect, is that the value it is supposed to correspond to is the file name of the extension (eg `0002-fixed-length-n-tuple-trees.md` or `0002-fixed-length-n-tuple-trees`) and that this should be clarified in the spec.

zimeon:
Fixes #477

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -976,23 +976,6 @@ DIGEST inventory.json
           <span id="E087">MUST</span> ignore any files in the storage root it does not understand.
         </p>
         <p>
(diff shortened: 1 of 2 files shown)


zimeon:
Cleanup remorse, closing....

PR #483, opened 2020-06-16
Define extension value as registered extension name

### LINKED ISSUE #477: Referencing extensions in ocfl_layout.json

zimeon:
Fixes #477 ... maybe ;-)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -976,29 +976,29 @@ DIGEST inventory.json
           <span id="E087">MUST</span> ignore any files in the storage root it does not understand.
         </p>
         <p>
-          An OCFL Storage Root MAY contain a JSON file named <code>ocfl_layout.json</code> to describe the arrangement
-          of directories and OCFL objects under the storage root. If present, this JSON document
-          <span id="E070">MUST</span> include the following two keys in the root JSON object:
-        </p>
-        <ul>
-          <li>
-            <code>key</code> - A key identifying the precise arrangement of directories and OCFL objects under the
-            storage root, i.e. how OCFL object identifiers are mapped to directory hierarchies. The value of this key
-            is not defined in the OCFL specification, but <span id="E071">MUST</span> correspond to a value given in an
-            <a>Extension</a>.
-          </li>
-          <li>
-            <code>description</code> - A human readable description of the arrangement of directories and OCFL
-            objects under the storage root.
-          </li>
-        </ul>
-        <p>
           An OCFL Storage Root <span id="E088">MUST NOT</span> contain directories or sub-directories other than as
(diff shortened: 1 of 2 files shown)


ahankinson:
Just flagging @pwinckles here so that he can see this.

draft/spec/index.html
```diff
@@ -975,30 +975,30 @@ <h2>Root Structure</h2>
...
         </p>
         <ul>
           <li>
-            <code>key</code> - A key identifying the precise arrangement of directories and OCFL objects under the
-            storage root, i.e. how OCFL object identifiers are mapped to directory hierarchies. The value of this key
-            is not defined in the OCFL specification, but <span id="E071">MUST</span> correspond to a value given in an
-            <a>Extension</a>.
+            <code>extension</code> - An extension name that identifies an arrangement of directories and OCFL objects
+            under the storage root, i.e. how OCFL object identifiers are mapped to directory hierarchies. The
+            value of the <code>extension</code> key <span id="E071">MUST</span> be the registered <a>extension</a>
```
awoods, line 994: There is still room for misinterpretation of what is meant by "registered extension name". I would suggest we refer to the proposed `Extension ID` seen in this pull-request: https://github.com/OCFL/extensions/pull/16/files#diff-77315547024f9923f13fb0a37b4ecf3aR3

zimeon: Or maybe we should fix that PR to say extension name? I'm loathe to call something that isn't a URI or similar an id

pwinckles: I don't mind updating that PR. I added that line speculatively.

zimeon: cf. "Digest Algorithm Name" also in https://ocfl.github.io/extensions/0001-digest-algorithms.html

awoods: Using a URI would be nice. 
It is important that the language here be very specific, unambiguous.

neilsjefferies: https://github.com/OCFL/extensions/pull/13 also has updates around extension names.

ahankinson:
Does that mean that this can be merged?

ahankinson:
Nevermind -- saw the comment in Slack

Issue #486, opened 2020-06-19
Clarify meaning of "registered extension name"


awoods:
Follow-on issue related to: https://github.com/OCFL/spec/issues/477

..see: https://github.com/OCFL/spec/pull/483#pullrequestreview-434083410

zimeon:
https://github.com/OCFL/extensions/pull/13/files attempts to do this with the following text:

> The *Registered Name* of an extension is the name of the extension file in the `docs` directory, excluding the `.md` extension.

If we want to do more _in the specification_ then I think a [terminology entry](https://ocfl.io/draft/spec/#terminology) for [registered extension name](https://ocfl.io/draft/spec/#dfn-registered-extension-name) would be the best approach, and it could then be linked from that term in both https://ocfl.io/draft/spec/#object-extensions and from https://ocfl.io/draft/spec/#root-structure

rosy1280:
I agree with @zimeon 's assessment and thumbs upped it to show agreement.  Would love to see if others have thoughts.  

cc'ing folks so we can get movement: @awoods @neilsjefferies @ahankinson @julianmorley

ahankinson:
I think we should have an entry in the terminology section, and will volunteer to write the PR if agreed.

ahankinson:
oops... @awoods beat me to it.

PR #487, opened 2020-06-19
Add definition of "Registered Extension Name"

awoods:
Resolves: https://github.com/OCFL/spec/issues/486

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -346,6 +346,13 @@
             The state of an <a>OCFL Object</a>'s content which is constructed using the incremental changes recorded in
             the sequence of corresponding and prior version directories.
         </dd>
+        <dt>
+            <dfn>Registered Extension Name</dfn>:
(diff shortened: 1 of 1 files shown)


ahankinson:
I note that @zimeon's comment on #486 says:

> The Registered Name of an extension is the name of the extension file in the docs directory, excluding the .md extension.

But it looks like the files also have a leading number? Is that part of the name too?

awoods:
@ahankinson : I think it is most clear if we use the "Extension Name" documented in the actual extension:
https://github.com/OCFL/extensions/pull/19/files#diff-10a1d32f04789570466246d3a20c4b6bR3

ahankinson:
So that does include the leading numbers. OK.

zimeon:
I think the numbers are part of it. I think this will all become very obvious when we have the **Extension Name: 1234-some-name** at the top of each extension

PR #491, opened 2020-07-03
1.0 Release

ahankinson:
Release 1.0 🎉 

This release contains a review of the respec parameters, and a custom build of respec with an OCFL profile.

The best way to review these changes is to look at the individual commits, since the 'Files Changed' view will just show you an aggregate view of the changes.

PR #495, opened 2020-07-07
Add date and links to implementation notes

zimeon:
Merge this into `release-1.0` before merging #491

I do wonder whether we should use `specStatus: "REC"` for the implementation notes instead of `"NOTE"` which then says this is a  "Working Group Note 07 July 2020" which simply isn't true. However, I haven't changed that

### DIFF
--- a/1.0/implementation-notes/index.html
+++ b/1.0/implementation-notes/index.html
@@ -8,8 +8,11 @@
             async class="remove"></script>
     <script class="remove">
         var respecConfig = {
+            publishDate: "2020-07-07",
             specStatus: "NOTE",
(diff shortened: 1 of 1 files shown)


ahankinson:
> specStatus: "REC" for the implementation notes instead of "NOTE" which then says this is a "Working Group Note 07 July 2020" which simply isn't true.

Given that the implementation notes are non-normative, I was hesitant to use the REC status, which seems to be the accepted status for official normative releases. I can also change what it says, so if you don't like "Working Group Note" I can change respec to generate an alternative.

zimeon:
Agree to changing "Working Group Note" in later fix

PR #506, opened 2020-07-30
New: Add a history section with original documents

ahankinson:
Perhaps it's a bit cutesy, but I also thought this might be an opportunity to save the files as an OCFL object as well. I know OCFL-in-git is a bit redundant, but in the interest of practicing what we preach, this might be a good, quick example that we can point to.

Thoughts and opinions welcome in the PR.


ahankinson:
Thanks @zimeon ! I used your OCFL Python library to create the inventory file, but got distracted between creating it and committing the PR. It should now be valid.

PR #552, opened 2021-05-25
Update version in text and examples to 1.1

### LINKED ISSUE #551: Update version in text and examples to 1.1
I think the default should be that the examples and text use the current OCFL specification version. We may wish also to create examples that include past version directories using prior versions of the specification

zimeon:
Fixes #551

### DIFF
--- a/draft/implementation-notes/index.html
+++ b/draft/implementation-notes/index.html
@@ -321,20 +321,20 @@
             OCFL Storage Roots).
             <pre>
             [storage_root]
-                ├── 0=ocfl_1.0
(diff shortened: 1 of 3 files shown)


rosy1280 (approved):

rosy1280:
Approved but wanted to note that if we want to create examples using past version numbers this PR doesn't do it.  I assume we would need to open a ticket for that?

zimeon:
Agreed. I thought I'd keep this in simple, easily approved, steps

awoods (approved):

PR #556, opened 2021-06-10
Create separate validation-codes.md for each version, fix links

### LINKED ISSUE #553: Should we version validation_codes.md?
I think when we created https://github.com/OCFL/spec/blob/main/validation/validation-codes.md we imagined consistency of error and warning codes between versions. I think we should certainly maintain this but I think we should version the validation codes file (perhaps move it into the spec directory) because:
  1. How do we add a new error code in draft or for a new version without creating confusion? E.g. https://github.com/OCFL/spec/pull/550
  2. Even if we maintain codes, we may change descriptions from version to version

zimeon:
Fixes #553


neilsjefferies (dismissed):

rosy1280 (approved):

awoods (commented):
We should also have a similar PR for `draft`, no?

zimeon:
Thanks @awoods -- I'd forgotten about the draft spec and draft implementation notes. Have fixes for those too.

I also noticed that the 1.0 implementation notes listed the implementation notes as an additional document. I fixed that to be the spec

awoods (approved):

neilsjefferies (approved):

rosy1280 (approved):

julianmorley (approved):

PR #560, opened 2021-09-21
Additional MUSTs for v version prefix and use of positive integers

### LINKED ISSUE #532: Missing requirement that non-zero-padded version numbers are prefixed with 'v'
Spec:

> The sequence of version numbers is the sequence of positive, base-ten integers: 1, 2, 3, etc., and the version directory name is constructed by adding the prefix v.

Validation codes:

- `E009`: ‘The version number sequence MUST start at 1 and must be continuous without missing integers.’
- `E010`: ‘The version number sequence must start at 1 and MUST be continuous without missing integers.’
- `E011`: ‘If zero-padded version directory numbers are used then they must start with the prefix v and then a zero.’

There is no code requiring non-zero-padded version numbers to begin with `v`. I assume this is because the spec is missing a MUST in the quoted sentence.

zimeon:
Fixes #532

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -451,10 +451,11 @@
         <h2>Version Directories</h2>
         <p>
             OCFL Object content <span id="E008">MUST</span> be stored as a sequence of one or more versions. Each
-            object version is stored in a version directory under the object root. The sequence of version numbers
-            is the sequence of positive, base-ten integers: 1, 2, 3, etc., and the version directory name is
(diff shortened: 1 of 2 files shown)


awoods (approved):

awoods:
Thanks, @zimeon .

neilsjefferies (approved):

rosy1280 (approved):

julianmorley (approved):

Issue #537, opened 2021-03-29
must a manifest entry be referenced in at least one version's state


pwinckles:
Fixture [E050_file_in_manifest_not_used](https://github.com/OCFL/fixtures/tree/main/1.0/bad-objects/E050_file_in_manifest_not_used), is a test for a manifest entry that is not referenced in any version state. It specifies the error should be `E050`:

> The keys of [the “state” JSON object] are digest values, each of which must correspond to an entry in the manifest of the inventory.

That text states that every logical path MUST have a corresponding manifest entry. It does not state that ever manifest entry MUST have at least one corresponding logical path. The nearest I can find to such a requirement in the spec is:

> The value of the manifest key is a JSON object, with keys corresponding to the digests of every content file in all versions of the OCFL Object.

But, that line, besides not having an associated validation code, is ambiguous. If "versions" are version objects in the inventory, then that could be read as requiring that every manifest entry be references in at least one version's state. But, if "versions" are version directories on disk, then it doesn't seem to be a requirement at all.

awoods:
Suggested resolution:
1. Update text of first sentence of [3.5.2 Manifest](https://ocfl.io/draft/spec/#manifest) to:
   > The value of the <code>manifest</code> key is a JSON object, all keys of which <span id="XXXX">MUST</span> correspond to digest value keys found in one or more state JSON objects of the current and/or previous <a>version</a> blocks of the <a>OCFL Object</a>.
2. Update [E050_file_in_manifest_not_used](https://github.com/OCFL/fixtures/tree/main/1.0/bad-objects/E050_file_in_manifest_not_used) to reference the newly created requirement code from above.
3. Create a new test that checks the scenario of [E050](https://ocfl.io/draft/spec/#E050) wherein a file in the version's state object is not found in the object's manifest.

zimeon:
Proposing a PR with minor wording change from the above:
> The value of the manifest key must be a JSON object, and each key must correspond to a digest value key found in one or more state JSON objects of the current and/or previous version blocks of the OCFL Object.

PR #561, opened 2021-09-22
Additional MUSTs re manifest keys

zimeon:
Part 1 of #537

(Will conflict with #560 so merge that first and then fix this)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -674,7 +674,7 @@
             <p>
                 The value of the <code>manifest</code> key <span id="E106">MUST</span> be a JSON object, and each key
                 <span id="E107">MUST</span> correspond to a digest value key found in one or more <code>state</code>
-                JSON objects of the current and/or previous <code>version</code> blocks of the <a>OCFL Object</a>. The
+                blocks of the current and/or previous <code>version</code> blocks of the <a>OCFL Object</a>. The
                 value for each key <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files
                 in the OCFL Object that have content with the given digest. As JSON keys are case sensitive, while
                 digests may not be, there is an additional requirement that each digest value
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -672,14 +672,15 @@
         <section id="manifest">
             <h2>Manifest</h2>
             <p>
-                The value of the <code>manifest</code> key is a JSON object, with keys corresponding to the digests
-                of every content file in the current and previous versions of the <a>OCFL Object</a>. The value for
-                each key <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the
-                OCFL Object that have content with the given digest. As JSON keys are case sensitive, while digests may
-                not be, there is an additional requirement that each digest value <span id="E096">MUST</span> occur
-                only once in the manifest regardless of case. Content paths within a manifest block
-                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>. The following restrictions
-                avoid ambiguity and provide path safety for clients processing the <code>manifest</code>.
+                The value of the <code>manifest</code> key <span id="E106">MUST</span> be a JSON object, and each key
+                <span id="E107">MUST</span> correspond to a digest value key found in one or more <code>state</code>
+                JSON objects of the current and/or previous <code>version</code> blocks of the <a>OCFL Object</a>. The
+                value for each key <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files
+                in the OCFL Object that have content with the given digest. As JSON keys are case sensitive, while
+                digests may not be, there is an additional requirement that each digest value
+                <span id="E096">MUST</span> occur only once in the manifest regardless of case. Content paths within a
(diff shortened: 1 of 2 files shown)


draft/spec/index.html
```diff
@@ -672,14 +672,15 @@ <h2>Basic Structure</h2>
...
-                of every content file in the current and previous versions of the <a>OCFL Object</a>. The value for
-                each key <span id="E092">MUST</span> be an array containing the <a>content path</a>s of files in the
-                OCFL Object that have content with the given digest. As JSON keys are case sensitive, while digests may
-                not be, there is an additional requirement that each digest value <span id="E096">MUST</span> occur
-                only once in the manifest regardless of case. Content paths within a manifest block
-                <span id="E042">MUST</span> be relative to the <a>OCFL Object Root</a>. The following restrictions
-                avoid ambiguity and provide path safety for clients processing the <code>manifest</code>.
+                The value of the <code>manifest</code> key <span id="E106">MUST</span> be a JSON object, and each key
+                <span id="E107">MUST</span> correspond to a digest value key found in one or more <code>state</code>
+                JSON objects of the current and/or previous <code>version</code> blocks of the <a>OCFL Object</a>. The
```
rosy1280, line 677: Do we call the state section a state JSON object in other locations?  I believe we generally refer to it as the state block.  I ask because the way this reads right now is a bit confusing and removing the `JSON objects` language might help that.

rosy1280: Also noting that this may have impact on the codes section.

zimeon: You are right, we say "`state` block". Will fix

zimeon: We are inconsistent about use of code font for "`xyx` block" -- will use it here, suggest overall fix https://github.com/OCFL/spec/issues/562

rosy1280 (dismissed):

neilsjefferies (dismissed):

draft/spec/validation-codes.md
```diff
@@ -113,6 +113,8 @@
 | E101 | 'Within an inventory, content paths must be unique and non-conflicting, so the content path for a file cannot appear as the initial part of another content path.' | https://ocfl.io/draft/spec/#E101
 | E102 | 'An inventory file must not contain keys that are not specified.' | https://ocfl.io/draft/spec/#E102
 | E103 | 'Each version directory within an OCFL Object MUST conform to either the same or a later OCFL specification version as the preceding version directory.' | https://ocfl.io/draft/spec/#E103
+| E106 | 'The value of the manifest key MUST be a JSON object.' | https://ocfl.io/draft/spec/#E106
+| E107 | 'The value of the manifest key must be a JSON object, all keys of which MUST correspond to digest value keys found in one or more state JSON objects of the current and/or previous version blocks of the OCFL Object.' | https://ocfl.io/draft/spec/#E107
```
awoods, line 117: A minor point, but the text of the error code is using the previous language:
> all keys of which MUST correspond..

instead of:
> and each key MUST correspond..

Aligning the language of the error code to the spec seems more even more clear.

awoods (changes requested):

awoods:
As expected, there is a (trivial) merge conflict.

draft/spec/validation-codes.md
zimeon: Well spotted @awoods ! I have updated the validation codes text to match the improved spec language

zimeon:
Have fixed merge conflict as well as addressing the error @awoods found

draft/spec/validation-codes.md
```diff
@@ -115,6 +115,8 @@
 | E103 | 'Each version directory within an OCFL Object MUST conform to either the same or a later OCFL specification version as the preceding version directory.' | https://ocfl.io/draft/spec/#E103
 | E104 | 'Version directory names MUST be constructed by prepending v to the version number.' | https://ocfl.io/draft/spec/#E104
 | E105 | 'The version number MUST be taken from the sequence of positive, base-ten integers: 1, 2, 3, etc.' | https://ocfl.io/draft/spec/#E105
+| E106 | 'The value of the manifest key MUST be a JSON object.' | https://ocfl.io/draft/spec/#E106
+| E107 | 'The value of the manifest key must be a JSON object, and each key MUST correspond to digest value keys found in one or more state JSON objects of the current and/or previous version blocks of the OCFL Object.' | https://ocfl.io/draft/spec/#E107
```
awoods, line 119: I don't want to be overly strict, but there are still minor differences in the text between the spec:
> The value of the manifest key MUST be a JSON object, and each key MUST correspond to a digest value key found in one or more state blocks of the current and/or previous version blocks of the OCFL Object.

..and the error code text:
> The value of the manifest key must be a JSON object, and each key MUST correspond to digest value keys found in one or more state JSON objects of the current and/or previous version blocks of the OCFL Object.

awoods (changes requested):

draft/spec/validation-codes.md
zimeon: Thanks again, fixed

awoods (approved):

neilsjefferies (approved):

rosy1280 (approved):

julianmorley (approved):

PR #576, opened 2022-02-04
Clarify that fixity value must be non-empty JSON object

### LINKED ISSUE #558: Do we need to clarify that `"fixity": null` is not allowed, or should we allow it?
From https://github.com/zimeon/ocfl-py/issues/79:
> @zimeon: I think it is clear in https://ocfl.io/1.0/spec/#fixity "The fixity block MUST contain keys corresponding to the controlled vocabulary given in the digest algorithms listed in the Digests section, or in a table given in an Extension." -- this implies a JSON object but if we want to add further clarification then that would be a spec issue
> @pwinckles: Right, but I would interpret that to mean that if there is a fixity block then it must have those properties. The fact that fixity is null, to me, means that it does not have a fixity block. Serializing missing fields to null is typical, and often configurable, for some JSON serializers.

zimeon:
Fixes #558

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -839,12 +839,13 @@
             <p>
                 An OCFL Object inventory MAY include a block for storing additional fixity information to supplement the
                 complete set of digests in the <a href="#manifest">Manifest</a>, for example to support legacy digests
-                from a content migration. This block <span id="E055">MUST</span> have the key of <code>fixity</code>
-                within the inventory.
+                from a content migration. If present, this block <span id="E055">MUST</span> have the key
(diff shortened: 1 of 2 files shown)


neilsjefferies (approved):

pwinckles:
If you insist with this change, see the issue for arguments against it, I would suggest that you also add a general statement to the [inventory section](https://ocfl.io/draft/spec/#inventory) that specifies your expectation that optional keys do not appear in the json with null values, which is a behavior statically typed json libraries will commonly do by default.

zimeon:
I'm going to close this. It is clear from continued discussion that we don't want to forbid both `null` and `{}`

Issue #557, opened 2021-07-02
Improve description of storage root extensions


zimeon:
We currently describe [Storage Root extensions](https://ocfl.io/1.0/spec/#storage-root-extensions) by reference to [Object extensions](https://ocfl.io/1.0/spec/#object-extensions). This is fine in terms of what the structure is and an error in the storage root extensions structure should result in `E086`. However, there is no warning code defined for use of an unregistered extension in a storage root. The code `W013` is used for object extensions but can't simply be used again for Storage Root extensions.

I think the best solution is simply to write out the structure requirement again for Storage Root extensions, commenting that this is the same as Object extensions. We can keep the error code `E086` and the define a new warning code for use of an unregistered extension.

awoods:
Two potential resolutions:
1. https://github.com/OCFL/spec/pull/583
2. https://github.com/OCFL/spec/pull/584

I lean towards the second.

zimeon:
Editors agree that the replication method provides a cleaner solution for readers and it is helpful to have separate error codes for the object and storage root cases.

zimeon:
The issue @pwinckles raised https://github.com/OCFL/spec/pull/583#pullrequestreview-935042817 about the proposed `E113` being unenforceable overlaps with #565. I think we could move forward with this issue by simply removing the text from the `E113` sentence onward in #583, and rely on a resolution of #565 to address the notion of what constitutes a registered extension name.

PR #582, opened 2022-04-07
Clarify that extension directory warning applies to Objects and the S…

awoods:
…torage Root.

Resolves: https://github.com/OCFL/spec/issues/557


zimeon:
Closing per editors' agreement https://github.com/OCFL/spec/issues/557#issuecomment-1091920032 on preference for alternative solution

Issue #565, opened 2021-10-08
Registered extension name clarification


pwinckles:
The spec defines an registered extension name as follows:

> The registered name of an extension is the name provided in the Extension Name property of the extension's definition. 

It later says:

> The specific structure and function of the extension, as well as a declaration of the registered extension name MUST be defined in one of the following locations:
> 
>   1. The OCFL Extensions repository
>   2. The Storage Root, as a plain text document directly in the Storage Root

The later seems to indicate that all extension names, adopted or not, are "registered" so long as there is a plaintext document that describes them in the storage root.

If this is true, the following codes present an impossible problem to validators because there is no way to know what is or isn't an extension name:

1. `E071`: ‘The value of the `ocfl_layout.json` extension key must be the registered extension name for the extension defining the arrangement under the storage root.’
2. `W013`: ‘In an OCFL Object, extension sub-directories SHOULD be named according to a registered extension name.’

I had originally thought that "registered" extension names were only for extensions that were adopted into the extensions repository. In which case, validators would enforce those rules for the set of known extensions. And, I was going to file a ticket to say that I thought that `E071` would be more congruous if it was a warning rather than an error.

So,

1. Are all extension names "registered"? If so, what are validators supposed to do?
2. If not, wouldn't it make more sense for `E071` to be a warning?

zimeon:
I agree that "registered" should mean "in the OCFL Extensions repository". I thus think that we should refine the definition of  [registered extension name](https://ocfl.io/1.0/spec/#dfn-registered-extension-name) as follows (_italic_ = added):

> **Registered Extension Name:**
> The registered name of an extension is the name provided in the Extension Name property of the extension's definition _in the [OCFL Extensions repository](https://ocfl.github.io/extensions/)_.

But, if we do that, how do we still leave space for **un**registered extensions? At present there is an underspecified notion of putting ["a plain text document directly in the Storage Root"](https://ocfl.io/1.0/spec/#object-extensions). To my mind two things are missing:

* A rational way to link this local extension via its extension name (the current unclear definition of "registered extension name" suggests that one might have to read every plain text document in the storage root and attempt to extract the name from the markdown?!?). I think we should instead just rely on the file name (perhaps minus extension)
* A naming convention for local extensions that avoids collision with registered extension names. That could be as simply as specifying that any local extension must have the prefix `xxxx-` or `local-` or something.

[...]

PR #583, opened 2022-04-07
Replicate specification of Object extension into Storage Root extensi…

awoods:
…ons section

Resolves: https://github.com/OCFL/spec/issues/557

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -1105,10 +1105,32 @@ DIGEST inventory.json
         </p>
         <p>
             The base directory of an OCFL Storage Root MAY contain a directory named <code>extensions</code> for the
-            purposes of extending the functionality of an OCFL Storage Root. The storage root <code>extensions</code>
-            directory <span id="E086">MUST</span> conform to the same guidelines and limitations as those defined for
+            purposes of extending the functionality of an OCFL Storage Root. The guidelines and limitations for the
+            storage root <code>extensions</code> directory are defined in alignment with those of the
             <a href="#object-extensions">object extensions</a>.
         </p>
+        <p>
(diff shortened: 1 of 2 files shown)


draft/spec/index.html
```diff
@@ -1105,10 +1105,32 @@ <h2>Storage Root Extensions</h2>
...
+            purposes of extending the functionality of an OCFL Storage Root. The guidelines and limitations for the
+            storage root <code>extensions</code> directory are defined in alignment with those of the
             <a href="#object-extensions">object extensions</a>.
         </p>
+        <p>
+            The <code>extensions</code> directory <span id="E112">MUST NOT</span> contain any files, and no
+            sub-directories other than extension sub-directories. Extension sub-directories
+            <span id="W016">SHOULD</span> be named according to a <a>registered extension name</a>. The specific
+            structure and function of the extension, as well as a declaration of the registered extension name
+            <span id="E113">MUST</span> be defined in one of the following locations:
```
pwinckles, line 1117: Personally, I would not include `E113` it is impossible to enforce.

zimeon: I agree that we should drop `E113` -- the "structure and function part" is not enforceable, and the notion of it being a registered name overlaps with the warning `W016`. (This also applies in the object extensions section for `E068` and `W013`.)

zimeon: I agree that we should drop `E113` -- the "structure and function part" is not enforceable, and the notion of it being a registered name overlaps with the warning `W016`. (This also applies in the object extensions section for `E068` and `W013`.)

zimeon: I note this has tons of overlap with https://github.com/OCFL/spec/issues/565, will propose approach there. I think we could move forward with this PR (addressing the replication issue alone) by simply remove the text from the `E113` sentence onward.

PR #584, opened 2022-04-07
Refine definition of registered extension name and describe local extensions

zimeon:
Attempt to address #565

(Does not include changes to [Storage Root Extensions](https://ocfl.io/draft/spec/#storage-root-extensions) that are being worked on in #557/#583, but should eventually match the approach for Object Extensions)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -293,7 +293,7 @@
         <dd>
             Extensions are used to collaborate, review, and publish additional non-normative functions related to OCFL.
             Extensions are intended to be informational and cite-able, but outside the scope of the normal specification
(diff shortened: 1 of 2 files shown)


pwinckles (approved):
👍 👍

rosy1280 (approved):

neilsjefferies (approved):

awoods (approved):
Look good. Thanks, @zimeon .
With this change in place, we can then align the text in [Storage Root Extensions](https://ocfl.io/draft/spec/#storage-root-extensions) and [Object Extensions](https://ocfl.io/draft/spec/#object-extensions)... as an iteration of https://github.com/OCFL/spec/pull/583 .

PR #586, opened 2022-04-18
Use the term 'logical state' consistently

### LINKED ISSUE #571: clarify "same object state" of version block (E066)
I have a question about this line from the spec: 

> `E066`: Each version block in each prior inventory file MUST represent the same object state as the corresponding version block in the current inventory file

As I understand it, the digest algorithm can change from one inventory to the next, which means the digests in the version blocks *can* change. If that's true, then isn't the sense of "sameness" in this statement somewhat ambiguous? I think it may help to explain what makes to version blocks the same, even when the digests may differ. Something like this: The same byte stream used for a given digest in the version block of the prior inventory must be used to generate the corresponding digest in the version block on the new inventory.

zimeon:
Fixes #571 and addresses two other instances where we said "object state" and should say [logical state](https://ocfl.io/draft/spec/#dfn-logical-state)

### DIFF
--- a/draft/spec/index.html
+++ b/draft/spec/index.html
@@ -108,7 +108,7 @@
             localBiblio: {
                 "BagIt": {
                     title: "The BagIt File Packaging Format (V1.0)",
-                    href: "https://tools.ietf.org/html/draft-kunze-bagit-17",
+                    href: "https://datatracker.ietf.org/doc/html/rfc8493",
                     authors: [
                         "J. Kunze",
                         "J. Littman",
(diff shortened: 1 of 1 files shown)


draft/spec/index.html
```diff
@@ -1353,9 +1353,10 @@ <h2>BagIt in an OCFL Object</h2>
         <p>
             [[BagIt]] is a common file packaging specification, but unlike the OCFL it does not provide a mechanism for
             content versioning. Using the OCFL it is possible to store a BagIt structure with content versioning, such
-            that when the object state is resolved, it creates a valid BagIt 'bag'. This example will illustrate one way
-            this can be accomplished, using the <a href="https://tools.ietf.org/html/draft-kunze-bagit-17#section-4.1">
-            example of a basic bag</a> given in the BagIt specification.
+            that when the <a>logical state</a> is resolved, it creates a valid BagIt 'bag'. This example will
+            illustrate one way this can be accomplished, using the
+            <a href="https://tools.ietf.org/html/draft-kunze-bagit-17#section-4.1"> example of a basic bag</a> given in
```
pwinckles, line 1358: Should this be https://datatracker.ietf.org/doc/html/rfc8493#section-4.1 ?

rosy1280 (changes requested):
Peter notes that the URL changed and we should update it to be good citizens.

draft/spec/index.html
zimeon: Thanks Peter!

neilsjefferies (approved):

rosy1280 (approved):

awoods (approved):

PR #601, opened 2022-06-03
Improve guidance on inclusion of markdown version of spec

### LINKED ISSUE #505: ocfl_1.0.txt
Congratulations on the 1.0 release!

The spec mentions:

> An OCFL Storage Root MAY contain other files as direct children. These might include a human-readable copy of the OCFL specification to make the storage root self-documenting

And gives `ocfl_1.0.txt` as an example. It would be a nice extra if this text document was generally available. I'm currently making it by hand.

I noticed that the ReSpec button at the top of the page offers a variety of export formats. Perhaps it could support txt too?

Or perhaps it's better to just put the HTML version in the storage root?

zimeon:
Fixes #505 

Now that we have the spec source in Markdown we can talk more explicitly about including that. I suggest a change to call it `ocfl_1.1.md` instead of using `.txt`, because the former is a more common extension for Markdown

### DIFF
--- a/draft/spec/index.md
+++ b/draft/spec/index.md
@@ -631,8 +631,10 @@ Declaration](#root-conformance-declaration) identifying it as such.
 
 An OCFL Storage Root <span class="rfc2119">MAY</span> contain other files as direct children. These might include a
 human-readable copy of the OCFL specification to make the storage root self-documenting, or files used to [document
-local extensions](#documenting-local-extensions). An OCFL validator <span id="E087" class="rfc2119">MUST</span> ignore
-any files in the storage root it does not understand.
+local extensions](#documenting-local-extensions). The source file for this specification document is in
+Markdown (described in \[[RFC7764](#ref-rfc7764)\], which is designed to be readable as plain text as well as for
+rendering as HTML, and thus makes it suitable for sefl-documentation. An OCFL validator <span id="E087"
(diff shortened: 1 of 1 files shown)


draft/spec/index.md
```diff
@@ -631,8 +631,10 @@ Declaration](#root-conformance-declaration) identifying it as such.
 
 An OCFL Storage Root <span class="rfc2119">MAY</span> contain other files as direct children. These might include a
 human-readable copy of the OCFL specification to make the storage root self-documenting, or files used to [document
-local extensions](#documenting-local-extensions). An OCFL validator <span id="E087" class="rfc2119">MUST</span> ignore
-any files in the storage root it does not understand.
+local extensions](#documenting-local-extensions). The source file for this specification document is in
+Markdown (described in \[[RFC7764](#ref-rfc7764)\], which is designed to be readable as plain text as well as for
+rendering as HTML, and thus makes it suitable for sefl-documentation. An OCFL validator <span id="E087"
```
julianmorley, line 636: typo

julianmorley (changes requested):
"sefl-documentation"

neilsjefferies (approved):

julianmorley (approved):

draft/spec/index.md
zimeon: Thanks, fixed

rosy1280 (approved):

PR #614, opened 2022-09-23
fix spelling

srerickson:
correct a spelling mistake.


zimeon (approved):

neilsjefferies (approved):

awoods (approved):

awoods (approved):
Thanks! Shall we also back-port the correction in 1.0?
https://github.com/OCFL/spec/blob/main/1.0/spec/index.html#L1123

zimeon:
I would say we leave 1.0 as-is. This spelling error would not lead to any misunderstanding even though it is nice to fix going forward

PR #640, opened 2023-09-23
List sections range

### LINKED ISSUE #637: Clarify that object structure is defined by sections 3.2 through 3.9
The comment in https://ocfl.io/1.1/spec/#object-structure:

> The [OCFL Object Root](https://ocfl.io/1.1/spec/#dfn-ocfl-object-root) MUST NOT contain files or directories other than those specified in the following sections.

is perhaps not as clear as it might be with regard to which the "following sections" are. We could make that explicit:

> The [OCFL Object Root](https://ocfl.io/1.1/spec/#dfn-ocfl-object-root) MUST NOT contain files or directories other than those specified in the following sections (3.2 through 3.9).

zimeon:
Fixes #637

### DIFF
--- a/draft/spec/index.md
+++ b/draft/spec/index.md
@@ -238,7 +238,7 @@ object validation. The structure for an object with one version is shown in the 
 ```
 
 The [OCFL Object Root](#dfn-ocfl-object-root) <span id="E001" class="rfc2119">MUST NOT</span> contain files or
-directories other than those specified in the following sections.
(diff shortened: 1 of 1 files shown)


neilsjefferies (approved):

rosy1280 (approved):

awoods (approved):
Section [3.4](https://ocfl.io/1.1/spec/#digests) is a bit of an outlier; but this PR is an improvement.

julianmorley (approved):

PR #641, opened 2023-09-23
Reword filesystem case sensitivity requirement

### LINKED ISSUE #528: Preserving filename case
The spec states:

> File paths and filenames in the OCFL are case sensitive. Filesystems MUST preserve the case of OCFL filepaths and filenames.

What exactly does the second sentence mean? Does it mean that case must be preserved within the inventory? Does it mean that case must be preserved on the filesystem?

Regardless of the intent, it's not clear to me what an implementation is supposed to do on filesystems like NTFS and HFS. They will technically preserve case, but they are also case insensitive.

What is a client supposed to do if `foo/bar.txt` and `Foo/Bar.txt` are inserted into an object? My assumption is that OCFL wants them to be treated as distinct logical paths, but how do you translate them to content paths? A direct logical path to content path translation would result in an unintended collision. It would seem that using OCFL on case insensitive filesystems would require a more invasive translation.

Similarly, I learned this evening that HFS additionally NFD normalizes all filenames, which may cause issues for OCFL content paths.

zimeon:
Fixes #528. Removes a MUST

### DIFF
--- a/draft/spec/index.md
+++ b/draft/spec/index.md
@@ -779,8 +779,8 @@ consistent manner. As such, the OCFL does not support the portability of these a
 OCFL Storage hierarchies. A common use case for links is storage deduplication. OCFL inventories provide a portable
 method of achieving the same effect by using digests to address content.
 
-3. File paths and filenames in the OCFL are case sensitive. Filesystems <span id="E091" class="rfc2119">MUST</span>
-preserve the case of OCFL filepaths and filenames.
+3. File paths and filenames in the OCFL are case sensitive. Implementations over filesystems that either do not preserve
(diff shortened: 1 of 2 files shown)


neilsjefferies (approved):

rosy1280 (approved):

awoods (changes requested):
The change looks good. In order to flag this change as one that impacts (reduces) the `MUST`s, it may be helpful to add that fact to the pull-request description... so that it shows up in the release notes.

zimeon:
@awoods - I amended the description to note that a MUST is removed

awoods (approved):

PR #662, opened 2024-11-20
Address digest table formatting (3.4)

awoods:

PR #663, opened 2024-11-21
Fix digests table in 1.1

zimeon:
Same as #662 for the live v1.1 (v1.1.1)


zimeon:
We will need to look at and work out what to do with the build test. At the moment it will always fail because of the long lines in the digests table in `draft/spec/index.md` (which is not changed by this PR)

neilsjefferies (approved):

rosy1280 (approved):

awoods (approved):

